Issue #5, opened 2020-08-05
`new` command


runemadsen:
We need to start building the Mechanic command-line client! 

There needs to be a `mechanic new` command that generates a project skeleton. This issue is about creating that command, but without the code that actually generates the project (we are just looking for the command-line flow). Here's what we imagined that the flow would be like. We want the experience to be really nice and clear, so take these instructions with a grain of salt and feel free to change anything that doesn't make sense to you.

```
$ mechanic new
$ Project name? (my-project)
   Select template for your first design function (you can always create more with `mechanic new function`)
> Vanilla JS (SVG)
   React (SVG)
   D3 (SVG)
   Vanilla JS (Canvas)
   p5.js (Canvas)
   THREE.js (Canvas)
$ Name your first function (my-function)
.. Generating project template
.. Installing dependencies
Done! Now run `cd my-project` and `npm run dev` 
```

A few notes on implementing this issue:

- Create a new `cli.js` file in the root of the `mechanic-cli` repo
- This file should use `yargs` to import a single command function for now, called `new` (you can place this in `commands/new.js`
- The command should use `inquirer` for the question prompts. We have also used `ora` (for a nice command-line spinner) and `chalk` (for coloring). Feel free to use these to create something nice
- Push your code to a `new-command` branch and make a PR to `master` when ready

You can see an example of this setup in [static](https://github.com/designsystemsinternational/clis/blob/master/packages/static/src/cli.js) and [dynamic](https://github.com/designsystemsinternational/clis/blob/master/packages/dynamic/src/cli.js) which use the same setup.

Once you have created a `new` command that prompts based on the questions above (or whatever you choose), we can start implementing the actual project generator.

Please ask any questions that you might have!

PR #6, opened 2020-08-05
New command

fdoflorenzano:
Added base of idea explained in #5 .

I took the opportunity to make a custom spinner to use with `ora` for DSI or maybe mechanic. That code was left in the `new command` script, but is probably best to separate in another file.


runemadsen:
This is fantastic @fdoflorenzano! I've never seen a cleaner commit history 👍

Issue #9, opened 2020-08-06
Generate UI based on mechanic parameters


runemadsen:
Each design function in a `mechanic` project has a `params` export that defines which parameters the function takes. Right now we only use the default `size` parameter:

```js
export const params = {
  size: {
    default: {
      width: 400,
      height: 300
    },
    large: {
      width: 1600,
      height: 1200
    }
  }
};
```

However, the first version of `mechanic` should allow users to add custom `string` and `integer` parameters like this:

```js
export const params = {
  //...  
  myLabel : {
    type: 'string'
  },
  myNumber: {
    type: 'integer'
  }
};
```

This should automatically create and string and number input field inside the [`Function`](https://github.com/designsystemsinternational/mechanic/blob/master/packages/mechanic-template/app/pages/Function.js) component using the [input components](https://github.com/designsystemsinternational/mechanic/tree/master/packages/mechanic-template/app/components/input).

Furthermore, it should be possible to add a `choices` attribute that restrict these values to a dropdown in the interface, as well as a `default` attribute that determines what the default value is: 

```js
export const params = {
  //...  
  myLabel : {
    type: 'string'
    choices: ["Rune", "Martin", "Esteban", "Ana", "Fernando"],
    default: "Rune"
  },
  myNumber: {
    type: 'integer',
    choices: [200, 300, 500],
    default: 200
  }
};
```

The selected values should then be passed along to the design function like the existing values. 

@fdoflorenzano, this is a slightly more complex issue, so please ask any questions that you might have. We can take a look at your code at any point if you are stuck.

PR #10, opened 2020-08-06
Generate ui based on params

fdoflorenzano:
Added parameter addition as described in #9 

All parameters are expected to have `default` and `type` properties, so that the UI and design function work. For the moment the supported types are `"string"` and `"integer"`, and if a `choices` property is present, then a select input is used instead of a regular HTML input.

- A new [`ParamInput`](https://github.com/designsystemsinternational/mechanic/blob/generate-ui-based-on-params/packages/mechanic-template/app/components/ParamInput.js) component was created at [`app/components`](https://github.com/designsystemsinternational/mechanic/tree/generate-ui-based-on-params/packages/mechanic-template/app/components), so that it's used by the `Function` component.

- [Validation](https://github.com/designsystemsinternational/mechanic/blob/7914f92d94d7cff54cb44a4c1aaad2bafe4f16fc/packages/mechanic-template/app/utils/mechanic.js#L22) for the supported properties was added.

- Also added custom properties for all working design functions as examples of use.

### DIFF
--- a/packages/mechanic-template/app/components/ParamInput.js
+++ b/packages/mechanic-template/app/components/ParamInput.js
@@ -4,14 +4,14 @@ import css from './ParamInput.css';
 import Input from './input/Input';
 import Select from './input/Select';
 
-export const ParamInput = ({ name, values, options, onChange, children }) => {
+export const ParamInput = ({ name, value, options, onChange, children }) => {
 
     const {type ,choices} = options;
     const _default = options["default"];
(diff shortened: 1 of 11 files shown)


runemadsen:
Fantastic! This looks spot on, so I'm merging this in now.

@fdoflorenzano Can you install `prettier` for your code editor and enable "Format on save" so the code auto formats?

Issue #11, opened 2020-08-07
Add boolean parameter


runemadsen:
I think we should include a `boolean` parameter type, alongside `string` and `integer`. For reference, I'm attaching an image of what the boolean toggle button looks like in our Flagship Pioneering tool, which I like a lot. I'm writing another issue about styling the toolbar in `mechanic-template`, but just pasting the images here for reference:

<img width="259" alt="Screenshot 2020-08-07 at 09 08 57" src="https://user-images.githubusercontent.com/192021/89619151-c7112780-d88d-11ea-872b-5eda964dd8ad.png">
<img width="262" alt="Screenshot 2020-08-07 at 09 08 50" src="https://user-images.githubusercontent.com/192021/89619157-c8daeb00-d88d-11ea-87db-33aed697a3fa.png">

PR #12, opened 2020-08-07
Boolean parameter

fdoflorenzano:
Added boolean parameter option, as requested in #11. The ui is a plain button to toggle it, expecting that styling will be revised on the toolbar as a whole in a different PR. Also made a pass over linting using prettier.


runemadsen:
Awesome. Looks great!

PR #14, opened 2020-08-07
:lipstick: Changed up botton component, toolbar structure and added CSS

fdoflorenzano:
Worked on Buttom component, the top function navigation section and tweaked the toolbar HTML organization.

This time there aren't so many commits, it kind of all came together at the end.


runemadsen:
This looks great! One ask before I merge this (and we can move on to the rest of the input fields): Can you use the [case](https://www.npmjs.com/package/case) package to turn the function names into capital case with spaces?

runemadsen:
This looks great! let's get it in, and go ahead with the buttons :)

PR #15, opened 2020-08-10
Input styling

fdoflorenzano:
Worked on the select, text and number inputs, made a shared base of appearance. 
Also re-organized (again) how stuff is rendered in the toolbar and added the scrolling section for parameters, as inspired by the Flagship Pioneering tool.
Added color to Export color to differentiate it from preview, using the logo blue.


runemadsen:
This is really nice! I think the line break for the labels is great, and the overall UI has a nice balance to it now.

@fdoflorenzano There is a bit of oddness when I try to change the input values. For example, I cannot delete `2` in `numberOfRects`. Can you take a look at that?

@bravomartin Take a look at this branch and add any comments that you might have!

bravomartin:
It's looking really nice!

My notes based on our conversation:

- change the function navigation into a special select where we can see all of the functions at once before selecting one.
- make size label the same as the others
- make props inputs the same width as size and buttons
- make labels smaller, use `Menlo, Courier New`.
- use same monospace in input value.
- make aside and root 100% height. 
- introduce a green for "true" in toggle.

fdoflorenzano:
Most of the proposed changes are there, I still want to make a pass at the whole CSS declaration and component structure.

runemadsen:
This looks great! Merging in.

PR #16, opened 2020-08-14
:bug: Patch fix for svg functions, adding SVG engine function

fdoflorenzano:
Fixed the error the SVG functions were throwing. Added a corresponding engine and adapted the element these functions returned to be more consistent as how the canvas functions did it. May need further refactoring of some details. This does not solve the video exporting error(which appears in all animation exporting).


packages/mechanic-template/app/utils/mechanic.js
```diff
@@ -254,6 +254,10 @@ export class Mechanic {
   async done(el) {
     if (!this.settings.animated) {
       if (isSVG(el)) {
+        // This conditional is a patch to an error, needs to be revised:
+        if (!this.serializer) {
+          this.serializer = new XMLSerializer();
+        }
```
runemadsen, line 260: Good point. I'm still not sure the `Mechanic` class structure is the best way to organize the shared mechanic functions, so let's discuss later today.

packages/mechanic-template/functions/index.js
```diff
@@ -93,12 +93,41 @@ const runP5 = (functionName, values, isPreview) => {
...
+  const func = functions[functionName];
+  const mechanic = new Mechanic(func.params, func.settings, values);
+
+  const onFrame = el => {
+    root.innerHTML = "";
+    root.appendChild(el);
+    if (!isPreview) {
+      mechanic.frame(el);
+    }
+  };
```
runemadsen, line 110: I knew this problem around appending a new SVG element on every frame would arise, and I'm afraid it will result in slow preview rendering times. We could consider using something along the lines of [virtual-dom](https://www.npmjs.com/package/virtual-dom) so we don't need to insert a new element on every frame.

packages/mechanic-template/functions/svgVideo/index.js
```diff
@@ -1,7 +1,9 @@
 export const handler = async (params, mechanic) => {
-  const r = params.width / 3;
+  const r = params.radius;
   let x = 0;
 
+  const template = document.createElement("template");
```
runemadsen, line 5: I hadn't considered the use of `<template />` here. I think it would be nice if function authors have the ability to just return an SVG string, so let's discuss this too.

fdoflorenzano:
This PR is still pending. I updated the structure to the current state of master so it complies to the engines as packages form. The virtual-dom implementation for the engine is the part is pending.

fdoflorenzano:
The virtual-dom implementation will be added on a later point.

PR #17, opened 2020-08-24
Package refactor

fdoflorenzano:
Separated engine definitions into packages, as well as the base mechanic utils used to require functions and define mechanic main class. This brought the need to adapt the template package, and got much simpler where it needed to be, as well as more explicit as what was happening.


runemadsen:
This looks great @fdoflorenzano! I think this is a really nice structure to start experimenting with the outputs. Let's talk about next steps!

Issue #19, opened 2020-09-07
To Do for v0.1.0 Release 


bravomartin:
- [x] Replace all default exports
- [x] Establish engine through require statement, import not really necessary.
- [x] Remove size param and replace with params  and presets. width and height are not special. Must have defaults.
- [x] “Fast-preview” →  “Scale to fit” (Works only for design functions that draw relative to the canvas size)
- [x] Separate mechanic input components into own package, so that unnecessary code is not added but still user has the option to modify it if wants.
- [x] Unique number input, that has slider option that turns it into a slider ui. min, max, step.
- [x] on Select: "choices" → “Options”,  add option for  `{ value: label }` objects as well as the current array of strings.
- [x] Add a nice color picker input.
- [x] Being able to remember params. Use localstorage.
- [x] Add validation function option, and add the possibility to show errors. Return null if no error, or a string if error.
- [x] Check that randomization is working
- [x] Fix animation export

fdoflorenzano:
Notes to self:

- [ ] Test new options declaration.
- [ ] Define a consistent way for mechanic to transform data types when returning values to df.
- [ ] Check CSS and search for now unused sections.

PR #20, opened 2020-09-07
Export rework

fdoflorenzano:
Did the first items for issue #19 
- Replace all default exports
- Establish engine through require statement, import not really necessary.

Also:
- Added prettier file to `mechanic-utils`
- Fixed little typo

### DIFF
--- a/packages/mechanic-engine-react/index.js
+++ b/packages/mechanic-engine-react/index.js
@@ -15,8 +15,7 @@ import React from "react";
 import { render, unmountComponentAtNode } from "react-dom";
 import { Mechanic } from "mechanic-utils";
 var root = document.getElementById("root");
-
-var run = function run(functionName, func, values, isPreview) {
+export var run = function run(functionName, func, values, isPreview) {
   unmountComponentAtNode(root);
   var mechanic = new Mechanic(func.params, func.settings, values);
   var Handler = func.handler;
@@ -62,9 +61,6 @@ var run = function run(functionName, func, values, isPreview) {
     done: onDone
   })), root);
 };
-
-var _default = run;
-export default _default;
 ;
 
 (function () {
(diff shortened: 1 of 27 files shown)


packages/mechanic-engine-react/index.js
```diff
@@ -74,9 +70,8 @@ export default _default;
     return;
   }
 
-  reactHotLoader.register(root, "root", "/Users/fdoflorenzano/Projects/DSI/mechanic/packages/mechanic-react-engine/src/index.js");
-  reactHotLoader.register(run, "run", "/Users/fdoflorenzano/Projects/DSI/mechanic/packages/mechanic-react-engine/src/index.js");
-  reactHotLoader.register(_default, "default", "/Users/fdoflorenzano/Projects/DSI/mechanic/packages/mechanic-react-engine/src/index.js");
+  reactHotLoader.register(root, "root", "/Users/fdoflorenzano/Projects/DSI/mechanic/packages/mechanic-engine-react/src/index.js");
+  reactHotLoader.register(run, "run", "/Users/fdoflorenzano/Projects/DSI/mechanic/packages/mechanic-engine-react/src/index.js");
```
runemadsen, line 74: Are these lines to make hot module loader work on development? We should probably figure out a way to do this without hardcoding the user path.

runemadsen (approved):
This looks great! Feel free to merge, but let's keep my one comment below in mind.

packages/mechanic-engine-react/index.js
fdoflorenzano: Nice observation. I didn't realize it was there since its the babel compilation output. That code is produced by the `react-hot-loader/babel` plugin, which I left in the `.babelrc` file as template from the template package. I read a little bit, and realized it not used by that package and I don't think it's needed. I removed it in the latest commit.

PR #21, opened 2020-09-07
Presets and size demotion

fdoflorenzano:
Did another two items in issue #19:
- Remove size param and replace with params and presets. width and height are not special. Must have defaults.
- “Fast-preview” → “Scale to fit” (Works only for design functions that draw relative to the canvas size)

I did it so that it always created a "default" preset option that sets every parameter into its default value, and also adds any option exported by the user in an `presets` object.

It also removes all "special" validation steps for size parameter. 

One aspect that wasn't discussed (I think) is the scaling effect now that the `width` and `height` parameters aren't always present. For now, it checks if those parameters are present, if they are and scaling down is active, then the passed down values are scaled. I was thinking that if scaling is active but one of the parameters are not present, then a bounding window object can be passed down. But I'm not sure.

### DIFF
--- a/packages/mechanic-engine-react/index.js
+++ b/packages/mechanic-engine-react/index.js
@@ -15,8 +15,7 @@ import React from "react";
 import { render, unmountComponentAtNode } from "react-dom";
 import { Mechanic } from "mechanic-utils";
 var root = document.getElementById("root");
-
-var run = function run(functionName, func, values, isPreview) {
+export var run = function run(functionName, func, values, isPreview) {
   unmountComponentAtNode(root);
   var mechanic = new Mechanic(func.params, func.settings, values);
   var Handler = func.handler;
@@ -62,9 +61,6 @@ var run = function run(functionName, func, values, isPreview) {
     done: onDone
   })), root);
 };
-
-var _default = run;
-export default _default;
 ;
 
 (function () {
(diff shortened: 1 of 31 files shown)


runemadsen:
This is great! Feel free to merge. 

About the `scaleToFit` if the `width` or `height` is not there: That is a good point, and I wonder if the scaling only works if those two are set? We can even just remove the option in the UI if both `width` and `height` are not set?

packages/mechanic-engine-react/index.js
```diff
@@ -74,9 +70,8 @@ export default _default;
     return;
   }
 
-  reactHotLoader.register(root, "root", "/Users/fdoflorenzano/Projects/DSI/mechanic/packages/mechanic-react-engine/src/index.js");
-  reactHotLoader.register(run, "run", "/Users/fdoflorenzano/Projects/DSI/mechanic/packages/mechanic-react-engine/src/index.js");
-  reactHotLoader.register(_default, "default", "/Users/fdoflorenzano/Projects/DSI/mechanic/packages/mechanic-react-engine/src/index.js");
+  reactHotLoader.register(root, "root", "/Users/fdoflorenzano/Projects/DSI/mechanic/packages/mechanic-engine-react/src/index.js");
```
bravomartin, line 73: shouldn't this url be relative to the project root?

runemadsen:
Ha ha, I commented this in the other PR

packages/mechanic-engine-react/index.js
fdoflorenzano: Commented on this issue on #20 😄

fdoflorenzano:
> This is great! Feel free to merge.
> 
> About the `scaleToFit` if the `width` or `height` is not there: That is a good point, and I wonder if the scaling only works if those two are set? We can even just remove the option in the UI if both `width` and `height` are not set?

That's a good option and simpler for us. Figuring out every possible way to scale it for every combination now seems like unnecessary. I went for the disabled button approach, cause the button disappearance without explanation seemed weird.

PR #22, opened 2020-09-21
More ui components and params

fdoflorenzano:
A lot of changes! v0.1.0 is closer!
Continued with items in issue #19:
- Separate mechanic input components into own package, so that unnecessary code is not added but still user has the option to modify it if wants.
  - Package is called `mechanic-ui-components`. It builds two folders, `src` that contains all components and `app` that contains a simple app that uses the components as examples.
  - `src` contains the `ParamInput` component that receives the attributes of a single parameter and works as a switch component that renders the corresponding input component based on the attributes.
  - `src` is further divided in `buttons` and `input` components. Now there's a single input component for ever type discussed: text, number, color, boolean and options. Each one works directly as an input or is a wrapper for an input and have very similar interfaces.
  - Removed two unused components: `Radio` and `Checkbox` since there have been no plans to use them. Then can be added back if we want, but their styling would be pending.
  - Removed all now unused styling and CSS from `mechanic-template`.
- Unique number input, that has slider option that turns it into a slider ui. min, max, step.
  - Added as `NumberInput`. Has inner logic to manage text or number props, and only propagates change in value if what is given can be parsed as a float.
  - Switchea between slider and number input if prop says so, and uses the min, max, step attributes.
- on Select: "choices" → “Options”, add option for { value: label } objects as well as the current array of strings.
  - Renamed attribute `choiced` to `options`.
  - Separated component into wrapper component `OptionInput` that uses `Select` but manages the logic of different types of option listings: either array of string or numbers, or an object.
  - If an object is passed, then the managed value in Select are the keys of the object, not the actual values. 
  - `mechanic-utils` was extended so that there is validation over this expected structure of options and in case of object listing, it passes the actual value selected to the design function instead of the key string. The latter allows for object value passing in options 😄.
- Add a nice color picker input.
  - Added as `ColorInput` that expects string values and returns string changes. Works as a button that when clicked a picker shows up to change the value.
  - Has special `model` prop that default to `rgba` that determines how the string will be returned on changes. Only tried `hex` and `rgba` since `hsla` was a little more complicated to parse nicely, but could be done.
  - Used [`react-color`](https://casesandberg.github.io/react-color/), specifically the `Chrome` component since it had variety of model selection and gave intuitive hue and transparency sliders.
- Being able to remember params. Use localstorage.
  - Used [`use-persisted-state`](https://github.com/donavon/use-persisted-state) in `mechanic-template` to define in state in local storage that separates params values by design function name.
- Add validation function option, and add the possibility to show errors. Return null if no error, or a string if error.
  - A parameter attribute `validation` can also be specified in a parameter for a design function. Must be a function that returns a string or null.
  - `ParamInput` component runs the given validation function if the current value changes, so that the returned result from the functions is shown with the component.
  - Styling for the ui error message could be revised.
  - Extended validation functions in `mechanic-utils` so that they also run the given `validation` functions when design functions is called.
  - Added validation for `validation` attribute in `mechanic-utils`, it just checks that it's a function.


runemadsen:
This is absolutely beautiful! Feel free to merge and let's talk about this in the meeting today.

PR #23, opened 2020-09-23
Export fix

fdoflorenzano:
Revised exporting issues as stated in #19. Three issues were detected:

- A simple one was an undefined attribute (`this.serializer`) in `mechanic` instances when `done` was called over an non-animated SVG image. It was fixed by adding conditional that checks if that attribute is undefined and creates new instance if so. This replicates other similar statements found in `frame` method. 
- Another simple issue was an not defined function calling (`isSVG`). That function was moved to separate file in previous refactor and wasn't properly renamed after.
- Video exporting for canvas isn't working on Firefox because of support for `webm-writer` library, specifically a canvas element method [(toDataURL)](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toDataURL). We decided that for now will only support Chrome-like browsers that support the necessary feature.
- SVG animation exporting was crushing because of missing size attributes needed to convert to canvas to then add as a frame to video buffer. Fixed by calling function that extracts size from SVG element on first frame calling. 

I think this fixes are fine for now, but the first mentioned issue makes me question if a refactor on the `mechanic` class flow could be done. Right now there is a lot of if statements and instance attributes that are used in some specific flows, so its hard to keep track what is going on. Also adding to the fact that some logic is passed to the engine, the limits between responsibilities is not completely clear to me yet.


runemadsen:
👍

PR #24, opened 2020-09-24
Ui components bug fixing

fdoflorenzano:
While passing the new structure to `dsi-logo-maker` and `shortcuts`, I noticed the following bugs that I've fixed:

- Changes on `NumberInput` were handled so that it did not update if the value wasn't valid, which produced that inputs could not be empty. Now if an empty string comes in (signal of empty input or invalid input), a 0 is added, which offers a better behavior.
- The label added in slider numeric inputs had variable length, which permitted that the actual slider width could change while sliding, making it difficult to use. Now that label has fixed width and does not overflow on big content.
- Missing validations in `mechanic-utils`, default value of a param could not me specified in options declared for that param. Added missing validation for both array and object versions.
- Locally stored values persisted between uses, including param names that change during usage and became unusable. I separated the value persistence handling into separate file and added filtering of params locally stored.

### DIFF
--- a/packages/mechanic-template/app/pages/Function.js
+++ b/packages/mechanic-template/app/pages/Function.js
@@ -3,8 +3,7 @@ import PropTypes from "prop-types";
 import classnames from "classnames";
 import Mousetrap from "mousetrap";
 
-import createPersistedState from "use-persisted-state";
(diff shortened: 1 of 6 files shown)


runemadsen:
👍 Feel free to merge!

PR #25, opened 2020-09-25
Dsi logo maker

fdoflorenzano:
Addition of a separate package that has same structure as `template` but filled with design functions that make different graphic resources using the DSI's logo identity. Code inside `functions` folder is as follows:

- `utils` has common code used to generate the logo's geometry and coloration. It builds logo "blocks" from character "bricks". It contains drawing functions for Canvas and React SVG components as well.
- `logoCanvas`  has a design function that creates a single logo block using a canvas element. Base size, word offset and color scheme can be altered.
- `logoSVG` same as `logoCanvas` but as SVG element.
- `logoAnimatedCanvas` same as `logoCanvas`  but animated word offset. Also duration of animation and number of loops that can be done can be changed.
- `logoAnimatedSVG` same as `logoAnimatedCanvas` but as SVG element.
- `fillCanvas` has a design function that fills a rectangular space with repetitions of the DSI logo using canvas element. Each repetition has a random flag color scheme and a different word offset. If a logo is cut at the end of a line, the missing portion appears at the beginning of the next line.
- `fillSVG` same as `fillCanvas` but as SVG element.
- `fillAnimatedCanvas` same as `fillCanvas` but animated word offset for each logo individually at a individual speed.
- `fillAnimatedSVG` same as `fillAnimatedCanvas`  but using SVG. Gets pretty laggy with many logos at once.
- `playCanvas` has a weird composition structure using different sized and offset-ed logo blocks using canvas element.
- `playSVG` same as `playCanvas` but as SVG element.
- `textCanvas` has a design function that creates a single logo block using a canvas element but with custom text and using canvas element. Base size, word offset, color scheme, number of columns and rows can be altered too.
- `textAnimatedCanvas` same as `textCanvas`  but that animated the word offset. Also duration of animation and number of loops that can be done can be changed.


runemadsen:
🏆 This looks great! I just have one question: Will it be annoying to maintain two versions of the same template?

fdoflorenzano:
> 🏆 This looks great! I just have one question: Will it be annoying to maintain two versions of the same template?

I think so too (it kinda already happened with the last big structural change), but I do not think it should go into template either. When we decide how to publish examples and stuff like that we can decide.

PR #26, opened 2020-09-25
:bug: Random seed is now remembered between preview and export calls

fdoflorenzano:
The current code in master had some code meant to manage random seed setting on mechanic's side. To work that seed had to be provided from the app when it tells the iframe to run the engine. The thing was that the app did not get a reference to any used seed, so it couldn't  remember any seed to give out on export calls.

When looking for old commits where this worked, I found [this one](https://github.com/designsystemsinternational/mechanic/commit/50125b2f7b1c606a90976ad25c6f684c94699d54). This did had a way to set the current random seed back to the app, through an event listener for an "init" event dispatched inside a `Mechanic` instance. The issue with this now is that `Mechanic`'s class structure and flow has changed since then and doesn't support that event (or any). 

This PR contains a way to give a reference of the used random seed back to the app, so it can successfully remember it on preview calls and pass it when export is called. For this, a couple main changes where needed:
- Since engine objects wrap the call of the `Mechanic` class where the value preparation is done (and the random seed setting), a way to get that value back in the app was needed. So now all engine functions return the created `Mechanic` instance after they render or call the design functions. 
- The code that runs in the iframe will receive the returned mechanic instance first, so here the value of the random seed is extracted to be returned back to the app. Returned `null` if no value es returned by the engine.
- The `app` in template remembers whatever the iframe returns when running the engine, and uses it for exporting if a value is there.

This adds a convention for engines. We can discuss if the best solution or not. It does open de possibility that more stuff is passed back to the iframe and app. Stats maybe? Warnings? Maybe just pass a selected group of objects back to the iframe and app. Let me know what you think.


packages/mechanic-template/app/pages/Function.js
```diff
@@ -50,7 +50,7 @@ export const Function = ({ name, exports, children }) => {
         height: bounds.height - 100
       };
     }
-    iframe.current.contentWindow.run(name, vals, true);
+    randomSeed.current = iframe.current.contentWindow.run(name, vals, true);
```
runemadsen, line 53: I would suggest, as you mention in your PR description, that we have a ref here called `lastRun` that stores the entire mechanic object. This makes it possible to do things like stats, etc. Then you would use `lastRun.values.randomSeed` to set the seed for the export.

runemadsen (changes requested):
This is some high quality digging :) Thanks for looking into this! I think this is a great solution, and I like the idea of being able to return stats / etc from the engines. I have suggested one small change.

packages/mechanic-template/app/pages/Function.js
fdoflorenzano: Done!

runemadsen: Thanks!

PR #27, opened 2021-07-05
Major package and organization refactor

fdoflorenzano:
Overall change on the previous setup we had. Now we approach it such that a mechanic project is a as minimal as possible, and a core `mechanic` dependency holds logic to setup server and launches an app with the current design functions in it.

Now, the entry point to mechanic is `create-mechanic`, which can be called using `npx` or the [`npm init initializer`](https://docs.npmjs.com/cli/v7/commands/npm-init) setup. The newest version is published in npm to test, so it can be used like so:
```
npm init @designsystemsinternational/mechanic@latest my-project
```

This will start a cli questionnaire with steps to set up a new project with a new design function inside. The function can be created based on a template or example, and can be passed directly as optional arguments:
```
# npm 6.x
npm init @designsystemsinternational/mechanic@latest my-project --template react
```
```
# npm 7+, extra double-dash is needed:
npm init @designsystemsinternational/mechanic@latest my-project -- --template react
```
```
# npm 6.x
npm init @designsystemsinternational/mechanic@latest my-project --example p5-animation
```
```
# npm 7+, extra double-dash is needed:
npm init @designsystemsinternational/mechanic@latest my-project -- --template p5-animation
```

Templates and examples are contained in `create-mechanic` for now, and hold some bases that we can work further in. 

Then, when inside a mechanic project, a couple commands from `mechanic-cli` are added as scripts in the `package.json`:
- `mechanic new function`: creates a new function inside the project.
- `mechanic dev`: serves the web app in place to try the design functions
- `mechanic build`: builds the code for web app and leaves the result in `dist/` directory by default
- `mechanic serve`: serves whatever was built using `build`.

`mechanic-cli` is really just a cli app that wraps functionality that `create-mechanic` or `mechanic` holds. 

`mechanic` holds commands related to serving and building the web app using webpack. It also has the app code, and the custom loader to inject the project design functions into the app.

`mechanic-ui-components` still holds the custom components for the app, it was simplified to not to a intermediate building to use in mechanic.

`mechanic-utils` holds some common cli definitions as colors, spinners and cli logos to be used by `create-mechanic`, `mechanic-cli` and `mechanic`.


runemadsen:
Looks great to me! Let's get it iiiiiiiiiiin

PR #28, opened 2021-07-07
Package Cleanup and fixes to some bugs

fdoflorenzano:
This PR adds a bunch of package configurations to all packages. In particular:
- `description` to all packages, so that they are available in the npm registry.
- `license` explicitly to all packages, lerna already linked it using the LICENSE in the root of the repository, but thought it would be better to be clear
- `homepage` to all packages, so that they are available in the npm registry. They all point to the main repo README for now, when we have a documentation site, we can change it to that.
- `repository` to all packages, so that they are available in the npm registry. They all point to the main repo but also state the directory to the specific package.
- `bugs` to all packages, they all point to the repo's issues page.
- `author` and `collaborators` to all packages. They all contain a permutation of Rune, Martín or me, depending on who started that specific package. All emails are `i@designsystems.international` and all url are `https://designsystems.international`
- `files` to all packages, so that only specific files are uploaded to the npm registry and we can ignore development files.
- `type` to most packages. Since the recent ESM support from Node, the recommended package configuration uses this field to specify if they are `"module"` (ESM) and `"commonjs"` (Common). They only one that doesn't have it explicitly is `mechanic` since it contains a mix of module uses.
- `exports` to most packages. Since the recent ESM support from Node, the recommended package configuration uses this field to specify entry points to the package instead of "main". 

Also used a similar order in all packages, following the this logic of field statements:
- Basics:
  - `name`
  - `version`
- Information:
  - `description`
  - `keywords` (not yet set)
  - `license`
  - `homepage`
  - `repository`
  - `bugs`
  - `author`
  - `contributors`
- Files:
  - `files`
  - `type`
  - `exports`
  - `module`
  - `bin`
- Tasks:
  - `scripts`
- Dependencies:
  - `dependencies`
  - `devDependencies`
  - `peerDependencies`
- Platform
  - `engines`
- Publishing
  - `private`
- Custom

I also added some basic READMEs to all packages. I re-used some of the one's i wrote for a previous version.

Also fixed some simple fixes found while testing.


runemadsen:
Absolutely brilliant!

runemadsen (approved):

PR #33, opened 2021-07-13
A bit of clean up

fdoflorenzano:
A spent some time going over some of the back mechanic functionalities, that I haven't touched in a while. I did some clean up in the core mechanic class and utilities and on the app.

- Added some comments on the function validation scripts
- Added some missing validation
- Refactored iframe code that loads engines and functions. Now it's not necessary to add the `.run` in engine definitions of functions.
- Updated examples and templates to fit this last change. Also fixed some new found bug in an example.
- Reordered a bit of the main `Function` component in the app. Added some optional chaining operation to handle edge cases, and added a "loading iframe" state so that buttons are disabled until they can really run engines and functions. This should fix the error @Anadroid got 🤞🏽 
- Added some DSI logo maker additions and changes from an older branch. Also adds the pin design function we used on December!


runemadsen (approved):
Beautiful!

PR #46, opened 2021-07-19
Editable params

### LINKED ISSUE #34: Editable params
Give parameters a way to be disabled based on current values. An `"editable"` field in param definition that takes a function that is evaluated on each render could be the way.

- [x] Adapt `ParamInput` component to expect `"editable"` field from parameters and evaluate it based of current values. If false, it disables the corresponding input.
- [x] Add styling to inputs to show they are currently disabled.
- [x] Add validation on mechanic core to check that `"editable"` are either functions or boolean.
- [x] Add examples in `dsi-logo-maker`

fdoflorenzano:
Give parameters a way to be disabled based on current values. An "editable" field in param definition that takes a function that is evaluated on each render could be the way. This resolves #34 and its task list.

You can check how the inputs look when disabled in a comment in #34, and check some examples in `dsi-logo-maker`:

https://user-images.githubusercontent.com/13511560/126200580-b9871f2c-e301-43c8-b5e9-35a37e3945e8.mov

This PR also:
- does some validation function refactor, improving code readability and error descriptions
- fixed some of the main README typos
- fixed some small accessibility in app: added title to HTML, [title to iframe](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#accessibility_concerns), and a label to design function select for navigation:

Screen Shot 2021-07-19 at 13 22 27

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Support for `"editable"` property in params. It disables the input in case of non editable result.
+
+### Changed
+
+- `ParamInput` component doesn't receive single `value` prop, but `values` prop with the whole object of values.
+
 ## [0.2.8] - 2021-07-15
 
 ### Changed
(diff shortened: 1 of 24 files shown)


runemadsen:
Very nice! I think this solution is really simple and flexible. It also make the tool seem a lot smarter, which is a bonus. Merge away 👍

runemadsen (approved):

PR #49, opened 2021-07-21
Add type to main mechanic package, engine requirements and some small fixes after testing

### LINKED ISSUE #41: Testing base on different environments
Linux:
- [x] Create basic project setup 
- [x] Run dev, build and serve
- [x] Add new design function
- [x] Use all different engines

Windows:
- [x] Create basic project setup
- [x] Run dev, build and serve
- [x] Add new design function
- [x] Use all different engines

Mac:
- [x] Create basic project setup
- [x] Run dev, build and serve
- [x] Add new design function
- [x] Use all different engines

Test out with different node versions:
- [x] 12.20.0
- [x] 14.13.1
- [x] 16.0.0

Bug issues:
- [x] #44
- [x] #48

### LINKED ISSUE #48: Add type to main `mechanic` package and engine requirements
To continue the ES Module adaption standard, the `mechanic` package should also state its type and adapt whatever necessary to work. Right now it also contains some Common JS scripts, but those don't need to be rewritten right now.

fdoflorenzano:
This PR:
- Resolves #48 
- Also contains some fixes noticed while doing #41. So merging this PR also resolves #41.
- Adds some new patch releases with this weeks updates.


runemadsen:
Awesome! I'm not sure I exactly understand why the imports now need `.js` and `.cjs`, but I trust you :)

fdoflorenzano:
> Awesome! I'm not sure I exactly understand why the imports now need `.js` and `.cjs`, but I trust you :)

Because when I add the `type` field to the package as `"module"`, node assumes its all ESM. If it tries to load a CJS script now without the explicit call out, it throws. So adding the `.cjs` tells the interpreter (?) what to expect and act accordingly.

PR #50, opened 2021-07-27
CSS Modules refactor to be compatible with Docs

fdoflorenzano:
This PR refactors and updates how CSS Modules where being imported into the react components for both `mechanic-ui-components` and the app in `mechanic`. So that it was also compatible with the way Gatsby loads these files. 
With this end, this PR:
- Updates `css-loader` and `style-loader`. We were using the previous major version of these, and the one we had for `css-loader` didn't support the option to use named exports.
- Changed the webpack configurations of both packages to enable [`namedExports`](https://webpack.js.org/loaders/css-loader/#namedexport) and updated the `exportLocalsConvention` option to `"camelCaseOnly"`. Before it had `"camelCase"` which was incompatible since it allowed exports that JavaScript can't define in an `import` statement.
- Changed all names of CSS Modules and added the `.module` extension before `.css`. Without this, Gatsby wouldn't detect exports.
- Changed all `import` statements of CSS to fit the updates.
- Published a new patch version of `mechanic` into npm with these updates.


runemadsen:
This makes me all happy and giggly inside. So awesome! Merge away!

runemadsen (approved):

PR #51, opened 2021-07-29
Image param

### LINKED ISSUE #35: Image param
New parameter type that gives user the ability to load images from their computer. Also, give options of predefined images loaded into the design function. The last part could be a separate param type from the first.

fdoflorenzano:
This PR adds an Image param! So this PR:
- Extended `ParamInput` to handle a the new `type == "image"` param and return the `ImageInput` component. 
- `ImageInput` component is also exported in `mechanic-ui-components`, as the rest of single inputs.
- `ImageInput` is the new input component that behind the scenes uses a file type input and hides it. A button is linked to the file's main function that starts a file browsing search. It only accepts image files. When loaded, all images are shown as little thumbnails of fixed height. Hover over images trigger a bigger preview of the image. A `multiple=true` property is expected that determines whether the input allows to load 1 or multiple images. Apart from all that, it follows a similar structure definition and styling as the other inputs in the library, but its a bit taller since space is needed for thumbnails and potentially a horizontal scrollbar (for many or long images). The value that it receives and also gives when a change is detected are  [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File) or [`FileList`](https://developer.mozilla.org/en-US/docs/Web/API/FileList) objects (depending on single or multiple param) that the file input element returns.
- Because we did not want to actually read and load these values into local storage (as the rest of the values thus far), I had to refactor and add and exception to the data serialization that just skips `File` and `FileLists` objects to be written into local storage. The rest of values for other inputs should still have persistance supported.
- Added into mechanic's core validation functions the cases needed for this new param.
- Added examples of the image param being used in the example app in `mechanic-ui-components` (can be tested with `npm run dev` in that package), that include single image, multiple image, disabled and "raw" styling.
- Refactored a bit the example app in `mechanic-ui-components` styling was defined.
- Adds two working design functions to `dsi-logo-maker` that use an image param. One based in Canvas and another in SVG.
- Also changed the reference name for F Grotesque that `dsi-logo-maker` had before from "F" to "F Grotesque Thin".
- Changed some grays being used in disabled inputs to accomplish full compliance for contrast accessibility. 

This resolves #35 

A couple things about it that maybe we want to revisit or discuss about this new param:
- Up until now, all params are required to have a `default` attribute to use as a initial value when a function is run for the first time. This can't be done with image since `File` and `FileList` objects are obtained through the file input interface. Right now validation just handles image params as an exception to the `default` attribute required rule. Should we revisit this rule or just keep the exception handling as is?
- Right now the value passed to design functions are `File` and `FileLists` objects, not really "images", which is kind of weird in concept. As such, its not that direct to use these objects in code to make them appear. It's not impossible, the examples in `dsi-logo-maker` are an example of that. Should the image reading and loading be left to the user or not would be the question here.


runemadsen (approved):
This looks great! About the image params not really being images, I think you have a good point: It would be great if we could handle the preloading of the image and then just pass the `Image` DOM element to the design function. Although, I could see people who wanted access to the core `Blob` object too. We could be inspired by what p5.js does with a [`preload` function](https://p5js.org/reference/#/p5/preload) where you can easily just preload images without needing the `onLoad` stuff. But let's get this in and consider that for the next update.

PR #57, opened 2021-08-10
Include fonts and styles on export

bravomartin:
This PR enables Mechanic to include fonts and styles in the exported SVGs. 
-  ensures function styles are inlined in production
-  inlines function fonts and css, and not mechanic's
-  improves optimizations to remove styles from other functions
-  adds ability to customize optimizations from the function settings
- works in react and vanilla svg templates


bravomartin:
@fdoflorenzano this is ready to review and merge

fdoflorenzano (approved):
Works on my end! Tried with a random otf and on the poster example using p5 and it all clicked! 
I added an section to the localIdentName of functions CSS cause without it I still got some confusion with same named CSS files for different folders.

Issue #59, opened 2021-08-12
Enable HMR for design functions


bravomartin:
**Is your feature request related to a problem? Please describe.**
The tool doesn't refresh when I make changes to my code

**Describe the solution you'd like**
I wolud like to fix hot module replacement so it works when functions are changed

**Describe alternatives you've considered**

**Additional context**
I see there is commented code that tries to do it and references to a bug.

there is this thread:
https://github.com/webpack/webpack-dev-server/issues/2792
but the [solution](https://github.com/webpack/webpack-dev-server/issues/2792#issuecomment-724169118) offered there didn't help. it might have to do with the fact that functions are in an iframe?

bravomartin:
💐

PR #65, opened 2021-08-23
Design function bundle refactor

fdoflorenzano:
This PR:
- Major refactor con how design functions are bundled. Now when calling the `dev` and `build` commands, it generates a temporary directory, the Mechanic project gets explored and looks for the root of each design function and for each one creates a new script that imports the design function. Each of these new temp scripts are the ones that get bundled into separate HTML files that then the main app will load in it's iframe. This simplifies the code that each iframe should run, and brings environment encapsulation between design functions for free. 
- Also does a refactor on language logged when running the main mechanic commands: `dev`, `build` and `serve`. They all print the logo at the beginning and end of the command, and they explain a bit further what they do.

I tried taking a stab with #59, but failed. Some remains of it are left in commits of this branch and other are left commented on the code. If @runemadsen wants to try, I would appreciate it. Can be in this same PR or another.


runemadsen (approved):
Beautiful! I think a separate bundle for each design function will be a huge relief moving forward. I can try to take a look at the hot module reloading some time this week. I'm not entirely sure where to start, but it must be possible!

PR #70, opened 2021-08-27
Iframe wrapping support

fdoflorenzano:
This PR adds a couple tweaks that enables the documentation project build and inject a mechanic project in an iframe.


runemadsen:
Beautiful! I really like how you solved this in a way that makes sense for the app in general, not just the documentation. Merge away!

PR #82, opened 2021-09-07
🐛 Fixed crush with used port

### LINKED ISSUE #81: Crush with used port.
**Describe the bug**
CLI commands `dev` and `serve` crush when trying to run with used port.

**Expected behavior**
Either show a message describing the issue and exiting, or actively use another port if one entered is used.

fdoflorenzano:
This resolves #81 . I tried using [get-post](https://www.npmjs.com/package/get-port), but [it didn't work as expected when giving it a port that is already in use](https://github.com/sindresorhus/get-port/issues/47). So I used [portfinder](https://www.npmjs.com/package/portfinder).


runemadsen (approved):
Nice! Approved!

PR #86, opened 2021-09-08
Create CLI updates

### LINKED ISSUE #84: Updates to create package
Right now the create package lists dependencies in a new project using the `latest` tag instead of an actual version, which is not the best. 

There is also the issue with running `npm init` with it and actually fetching the latest version.

fdoflorenzano:
This PR includes changes to the project creation CLI for mechanic. These are:
- The biggest change, is the renaming and moving of the package. The package is now called `create-mechanic` instead of `@mechanic-design/create`. This lets a user use `npm init mechanic@latest` without issues. In the repository, everything was moved to a new package folder `create-mechanic`, and the `create` folder (that points to the now deprecated  `@mechanic-design/create`) was reduced to a minimal console log script that let's the user know that they should use  `create-mechanic`. I left the package folder just so there was still a place to refer to it even if its deprecated. After a last release of it, I made it private so that updates in the monorepo are not released.
- One functionality update added to the CLI is a script (`packages/create-mechanic/checkVersion.js`) that runs when the CLI is called before running the main create functionality. This script check that the running version of the package isn't older that the current release made in the npm registry. If it is, it logs a warning letting the user know they are running an older version. This would be helpful in case a user runs  `npm init mechanic` (without `@latest`) and an older version that got cached locally is used.
- Another functionality added is the addition or a pre-publishing script (`packages/create-mechanic/updateVersions.js`) that updates the versions of all dependencies of templates and examples. That way, when running publish through `lerna`, this script will update all dependencies to the newest version of each package, and that way the CLI uses update versions in the projects (instead of `latest`).

This resolves #84


runemadsen (approved):
Very nice! I never realized that the `init` command could use a locally cached version of the package. So I guess me running `npm init gatsby` would use a cached version if I had `gatsby` installed as a global package? Anyway, this is great 👍

fdoflorenzano:
> Very nice! I never realized that the `init` command could use a locally cached version of the package. So I guess me running `npm init gatsby` would use a cached version if I had `gatsby` installed as a global package? Anyway, this is great 👍

Could be! There's definitely some weird behaviors with it. The [docs](https://docs.npmjs.com/cli/v7/commands/npm-init) say it uses `npm exec`, which supposedly checks local install. But I'm not sure it checks global install. But maybe with the `@latest` tag you can force it.

PR #96, opened 2021-09-15
Change `usesRandom`

### LINKED ISSUE #68: Use seeded random by default

When I'm using a design function and decide to use random on it, it is unsettling that the version changes when I click export. 
I would expect it to "just work".
Persisting the random from preview to export is the expected behavior, and it doesn't do any damage or add significant overhead that it is on all the time even if you're not using it. 
There might be edge cases where you don't want it, for which case we can leave it as a setting that you can disable, similar to what we did for `optimize`.

- rename `usesRandom` to `persistRandomOnExport`.
- default to `true`

fdoflorenzano:
This resolves #68


runemadsen:
Beautifuuuuuul!

PR #97, opened 2021-09-15
Add font to loading page

### LINKED ISSUE #75: Add Object Sans fontface to loading page
The current loading page doesn't have any Mechanic specific font-face. We should fix that eventually. I gave it a go in the ` add-font-to-loading-page` branch without success.

fdoflorenzano:
This resolves #75 . It does so by also serving the font from the local server temporarily so that the page can use it.


bravomartin (approved):
looks great!

PR #103, opened 2021-09-29
Added new examples and new react-image template

fdoflorenzano:
This PR adds to `create-mechanic` multiple code bases an their corresponding dependency files:
- A new and simpler template for the `react-image` option.
- A new business card generator example
- A new Instagram story generator example

After this is merged, a new minor release will be published.


runemadsen (approved):
Beautiful! Approved 👍

Issue #105, opened 2021-10-06
Add html-dom support


bravomartin:
The currently available engines can only handle `svg` or `canvas` elements.

What if we added an html-engine that can take any dom node as the root? This would make it much easier to create text-heavy design functions, and would unlock many layout features unavailable to svg.

There is [dom-to-svg](https://www.npmjs.com/package/dom-to-svg) that seems to work very well. [This](https://github.com/felixfbecker/svg-screenshots) chrome / firefox extension uses it and it seems to work great

We should of course take a solid look at the implications, etc.

PR #113, opened 2022-02-08
Export HTMLElement in React to image and video

caroillemann:
When exporting dom-elements that use pixel values in styling, you need to account for the scale factor so the sizes of all elements look the same in both the full-size preview and the downscaled preview. Therefore I had to expose the scale ratio along with the original width and height to the handler in the 'inputs' prop, so the handler can account for that.

### DIFF
--- a/packages/core/app/pages/Function.js
+++ b/packages/core/app/pages/Function.js
@@ -34,20 +34,12 @@ export const Function = ({ name, exports: functionExports, children }) => {
 
   const prepareValues = (useScale, useRandomSeed) => {
     const valuesCopy = Object.assign({}, values);
-    valuesCopy.scaleToFit = {
-      widthOriginal: values.width,
-      heightOriginal: values.height,
-      width: values.width,
-      height: values.height,
-      factor: 1,
-    };
     if (useScale && canScale && scaleToFit) {
       const bounds = mainRef.current.getBoundingClientRect();
-      const availableWidth = bounds.width - 100
-      const availableHeight = bounds.height - 100
-      valuesCopy.scaleToFit.width = availableWidth < values.width ? availableWidth : values.width;
-      valuesCopy.scaleToFit.height = availableHeight < values.height ? availableHeight : values.height;
-      valuesCopy.scaleToFit.factor = valuesCopy.scaleToFit.width / values.width
+      valuesCopy.scaleToFit = {
+        width: bounds.width - 100,
+        height: bounds.height - 100
+      };
     }
     if (useRandomSeed && lastRun.current?.values) {
       valuesCopy.randomSeed = lastRun.current.values.randomSeed;
--- a/packages/core/app/pages/Function.js
+++ b/packages/core/app/pages/Function.js
@@ -34,17 +34,30 @@ export const Function = ({ name, exports: functionExports, children }) => {
 
   const prepareValues = (useScale, useRandomSeed) => {
     const valuesCopy = Object.assign({}, values);
+    valuesCopy.scaleToFit = {
+      widthOriginal: values.width,
+      heightOriginal: values.height,
+      width: values.width,
+      height: values.height,
+      factor: 1,
+    };
     if (useScale && canScale && scaleToFit) {
       const bounds = mainRef.current.getBoundingClientRect();
-      valuesCopy.scaleToFit = {
-        width: bounds.width - 100,
-        height: bounds.height - 100
-      };
+      const availableWidth = bounds.width - 100
+      const availableHeight = bounds.height - 100
+      valuesCopy.scaleToFit.width = availableWidth < values.width ? availableWidth : values.width;
+      valuesCopy.scaleToFit.height = availableHeight < values.height ? availableHeight : values.height;
+      valuesCopy.scaleToFit.factor = valuesCopy.scaleToFit.width / values.width
     }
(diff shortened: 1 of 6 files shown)


fdoflorenzano:
Hey @caroillemann! Thank you so much for doing this, we really appreciate it! We hope Mechanic has helped you in some way! 

I was a little bit confused at first cause the title and description kinda talk about different things. I'm guessing one thing is the support of exporting HTMLElement as images and video, and another is the inclusion of the ratio as an input value to the functions handler. 

Could you describe the use case of exporting of HTMLElement? Just for me to really take into account what you are trying to do and if there is anything missing. Maybe also adding a design function example would be helpful. It doesn't have to be added to the project, just as a comment here is enough :) 

Thanks again!

packages/core/app/pages/Function.js
```diff
@@ -47,6 +47,11 @@ export const Function = ({ name, exports: functionExports, children }) => {
     return valuesCopy;
   };
 
+  const handleScaleToFit = () => {
+    setScaleToFit(!scaleToFit);
+    iframe.current.contentWindow?.dispatchEvent?.(new CustomEvent('scaleToFit', { detail: { scaleToFit: !scaleToFit } }));
```
fdoflorenzano, line 52: Is this line necessary? Since the state will be updated, then the preview should automatically also trigger. Or maybe that's not why this is here. Let me know!

packages/core/src/mechanic-validation.js
```diff
@@ -197,16 +196,24 @@ const prepareValues = (inputs, settings, baseValues) => {
...
-      : 1;
-    if (ratioWidth < 1 || ratioHeight < 1) {
-      const ratio = ratioWidth < ratioHeight ? ratioWidth : ratioHeight;
-      values.width = Math.floor(values.width * ratio);
-      values.height = Math.floor(values.height * ratio);
+  // Add ratio and original values if width and height are inputs
+  if (values.width && values.height) {
+    values.widthOriginal = values.width
+    values.heightOriginal = values.height
+    values.ratio = 1
```
fdoflorenzano, line 203: I'm thinking that maybe the names for this could be different, just so that there's no name collision with the inputs defined by the user. Even though `widthOriginal` and `heightOriginal` don't seem like go-to names, `ratio` does and I've even used it. 

Maybe prepending an underscore to all "added" values from Mechanic could be the way. So `_width`, `_height` and `_ratio`. I'm not totally convinced about that specific naming convention, so I would love to hear your thoughts on it!

packages/core/src/mechanic-validation.js
```diff
@@ -218,15 +225,21 @@ const prepareValues = (inputs, settings, baseValues) => {
...
+ */
+const isCanvas = el => el instanceof HTMLCanvasElement;
+
 /**
  * Validates that a DOM element is SVG or Canvas
  * @param {object} el - A DOM element to check
  */
 const validateEl = el => {
-  if (isSVG(el) || el instanceof HTMLCanvasElement) {
+  if (isSVG(el) || isCanvas(el) || el instanceof HTMLElement) {
```
fdoflorenzano, line 239: Just out of curiosity, aren't canvas elements also recognized as HTMLElement?

packages/core/src/mechanic.js
```diff
@@ -38,7 +40,8 @@ export class Mechanic {
       throw new MechanicError(err3);
     }
 
-    this.inputs = inputs;
+    // this.inputs = inputs;
+    this.inputs = validation.prepareValues(inputs, settings, values);
```
fdoflorenzano, line 44: Is this necessary or accurate?

fdoflorenzano (commented):
I've added some comments and questions for you about some of the details in the implementation.

I tested locally with the provided example and it does seem to work! When the comments are reviewed, we'll try to make further testing of other examples so we are sure everything works well.

So just for you to know, we are currently working on some feature additions that affect the files that you are working on, so depending on when we merge each thing, we would have to figure out conflicts. But I doubt our additions will be added soon. 

Thank you so much for this again! It's exciting for us to see. This potentially would resolve #105.

packages/core/app/pages/Function.js
caroillemann: Agree - it's not necessary when the scale to fit ratio is included in the state. I've removed it.

packages/core/src/mechanic-validation.js
caroillemann: I think it's a good idea to prefix it - and an underscore sounds fine. I've changed it in commit [19211bc](https://github.com/designsystemsinternational/mechanic/pull/113/commits/19211bce4ad9cd54e93590b69c124a9581ea4599).

packages/core/src/mechanic-validation.js
caroillemann: True - have been updated in [19211bc](https://github.com/designsystemsinternational/mechanic/pull/113/commits/19211bce4ad9cd54e93590b69c124a9581ea4599).

fdoflorenzano (approved):
This is great! I spent some time reviewing and testing more cases and it all seems to be working well! Thanks for the changes and again for the PR! 

Doing this I realized it's been quite a while since the last release, so maybe we'll publish one soon with the unreleased changes. Now including this!

caroillemann:
Perfect - yes, that would be great 😊

Issue #116, opened 2022-03-18
Error when providing options - "values is not defined"


reaganchisholm:
**Describe the bug**

```
ReferenceError: values is not defined - You may find more context in console.
```

**To Reproduce**
1. Add options to input to convert it to select field type

```js
export const inputs = {
  color: {
    type: "number",
    default: 1,
    options: [1, 5, 10], 
  },
  // ...
}
```

**Expected behavior**
Select dropdown should be provided with options without rendering error.

**Screenshots**
![image](https://user-images.githubusercontent.com/19511364/158927502-432f40c8-f9eb-4352-a969-53dd2919d2a6.png)

**Desktop**
 - Windows
 - Chrome
 - 2.0.0-beta.1

**Additional context**
I tested this out on version `1.2.0` and it works as expected so it seems related to most recent `2.0.0-beta` release.

fdoflorenzano:
Thanks for this! We'll try to fix it as soon as possible.

fdoflorenzano:
Hey @reaganchisholm! We just released `v2.0.0-beta.2`, and I think you'll find that the mentioned bug doesn't occur anymore.

reaganchisholm:
Working as expected now, thanks!

PR #117, opened 2022-03-18
Fixes for number options input validation and error catching in design function execution

fdoflorenzano:
This PR fixes a couple bugs.

First, it fixes a typo in the code that validates an options input (`values` => `value`) that was noted thanks to issue #116. 

By reviewing that, I realized that validation over a number input that has options provided also would arrive to an error because the select input used returns a string and not a number, so I added some extra sentences that handles that case.

On top of that, I realized that validation over the design function's inputs showed errors correctly, but any errors that appeared when validation the values and running the design function broke the app and it's rendering. This was because the error throwing permeated into the React App and didn't just throw in the iframe. I added some return values back to the app that would correctly be received in the app, and set the error throw in a timeout so that it stays in the iframe.

If this is merged, we would release version `v2.0.0-beta.2`.

### DIFF
--- a/packages/core/src/base-inputs/number.js
+++ b/packages/core/src/base-inputs/number.js
@@ -21,27 +21,35 @@ export default {
   },
   requiredProperties: ["default"],
   validation: (inputValue, input) => {
-    if (typeof inputValue !== "number") {
+    if (typeof inputValue !== "number" && typeof inputValue !== "string") {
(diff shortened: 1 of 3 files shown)


runemadsen:
Looks good to me! Good catch on the string/int functionality. I think this is a nice little thing to handle automatically. Merge and publish away!

PR #120, opened 2022-03-22
Nicer Generate Button

bravomartin:
- Renamed Preview / Randomize to be `Generate` 
- Changed the styles for the prev and next buttons to be sub-actions of the Generate button

This required some changes to the core UI library:
- Updated how disabled works to prevent disabled buttons to trigger the button below
- Made changes to ensure the disabled buttons are not focused

https://user-images.githubusercontent.com/605285/159568359-5ead334f-ee6c-4955-b3b7-c0ebc7883a8f.mov

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -72,9 +72,9 @@
 			}
 		},
 		"node_modules/@ampproject/remapping": {
-			"version": "2.1.1",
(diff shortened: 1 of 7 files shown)


fdoflorenzano (changes requested):
Works and looks great! The only missing thing I would ask you is adding a CHANGELOG record for the `ui-components` change.

fdoflorenzano (approved):
🏅

PR #122, opened 2022-04-10
Slider input UI improvements

### LINKED ISSUE #121: Slider input element value not visible properly when zero
**Describe the bug**
For a slider input element if the number is 0 it's not visible completely and there are 3 dots (view screenshots)

**To Reproduce**
1. Create a slider input element
2. See error

**Expected behavior**
The whole number should be visible and no dots

**Screenshots**
Screenshot 2022-03-24 at 17 05 35
Screenshot 2022-03-24 at 16 55 11

**Desktop (please complete the following information):**
 - MacOS 12.2.1
 - Brave 1.35.104
 - Mechanic 1.1.0

lnolte:
Fixes #121 

Changes the `flex-basis` for the slider value's container from an `em` value to a `ch` value. `1ch` is measured as the width of the current font's zero, so it lends itself to be used for estimating the width of a mostly numeric string.

Also changes the value to a fixed length string based on the number of decimal digits, which makes the UI feel a little less "jumpy" when changing slider values.

### DIFF
--- a/packages/ui-components/CHANGELOG.md
+++ b/packages/ui-components/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- Updated width of number value display to `ch` units in slider so it does not overflow
+
 ## 2.0.0-beta - 2022-03-24
 
 ### Added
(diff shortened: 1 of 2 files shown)


fdoflorenzano (changes requested):
That's amazing @lnolte! Thank you so much for this. I tried it locally and it looks great, and the changes you made make sense to me.

The only think I would ask you to add is a record in the `ui-components/` CHANGELOG file adding a description of the change. With this and merging I can publish a little patch version.

Thanks again!

fdoflorenzano (approved):
Amazing! Thanks again for this!

PR #127, opened 2022-05-03
Fixes filename format

bravomartin:
Changed filename from `[name][timestamp].[ext]` to `[filename]-[timestamp].[ext]` for better readability


fdoflorenzano (approved):
Nice and simple!

PR #131, opened 2022-05-09
DSI Flag Builder

lnolte:
Design Functions (SVG and Canvas) that allow building deterministic flags from Text input. This is used on the current DSI website to generate the placeholder images covering loading images.

It would also be nice to have this available as a design function to build flags for other purposes (we have a soccer world cup coming up this year after all … just kidding).


fdoflorenzano (approved):
I love it!

PR #134, opened 2022-06-07
D3 template added

munusshih:


lerna.json
```diff
@@ -2,5 +2,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "2.0.0-beta.7"
+  "version": "2.0.0"
```
fdoflorenzano, line 5: Why the change to version `2.0.0`? As far as I'm aware we haven't updated to this release. Also, isn't this PR for a D3 template? Why is this included?

packages/cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mechanic-design/cli",
-  "version": "2.0.0-beta.7",
+  "version": "2.0.0",
```
fdoflorenzano, line 3: Same comment about new release.

packages/cli/package.json
```diff
@@ -43,7 +43,7 @@
   },
   "dependencies": {
     "@mechanic-design/utils": "^1.1.0",
-    "create-mechanic": "^2.0.0-beta.7",
+    "create-mechanic": "^2.0.0",
```
fdoflorenzano, line 46: Same comment about new release.

packages/create-mechanic/function-templates/d3-image/dependencies.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "@mechanic-design/engine-svg": "^2.0.0-beta.7"
+  }
+}
```
fdoflorenzano, line 5: Why isn't D3 included as a dependency for this template?

packages/create-mechanic/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "create-mechanic",
-  "version": "2.0.0-beta.7",
+  "version": "2.0.0",
```
fdoflorenzano, line 3: Same comment about new release.

packages/dsi-logo-maker/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mechanic-design/dsi-logo-maker",
-  "version": "2.0.0-beta.7",
+  "version": "2.0.0",
```
fdoflorenzano, line 3: Same comment about new release.

fdoflorenzano (changes requested):
I left a first batch of comments, mostly regarding some confusing additions. 

Also, I would appreciate you adding a description to the PR so I get some context of what it's supposed to be added in it. Maybe some of the questions I added could have been answered if a description was included.

lerna.json
munusshih: I'm also confused... I'll delete these parts then. I don't think I've touched these parts of the code.

PR #135, opened 2022-06-08
Adding D3 and SVG.js Templates

munusshih:
- Add a `d3-image` template to the `function-template` folder
- Add `d3` to the dependency of the template
- Add a `svgjs-image` template to the `function-template` folder
- Add `svg.js` to the dependency of the template
- Add both `d3-image` and `svgjs-image` to the `index.js` of `function-template`

### DIFF
--- a/packages/create-mechanic/function-templates/d3-image/dependencies.json
+++ b/packages/create-mechanic/function-templates/d3-image/dependencies.json
@@ -1,6 +1,6 @@
 {
   "dependencies": {
     "@mechanic-design/engine-svg": "^2.0.0-beta.7",
-    "d3": "^7.4.4"
+    "d3": "^7.4.4",
   }
 }
--- a/packages/create-mechanic/function-templates/d3-image/dependencies.json
+++ b/packages/create-mechanic/function-templates/d3-image/dependencies.json
@@ -1,6 +1,6 @@
 {
   "dependencies": {
     "@mechanic-design/engine-svg": "^2.0.0-beta.7",
-    "d3": "^7.4.4",
+    "d3": "^7.4.4"
   }
 }
--- a/packages/create-mechanic/function-templates/d3-image/dependencies.json
+++ b/packages/create-mechanic/function-templates/d3-image/dependencies.json
@@ -1,0 +1,6 @@
+{
+  "dependencies": {
+    "@mechanic-design/engine-svg": "^2.0.0-beta.7",
+    "d3": "^7.4.4"
+  }
+}
--- a/packages/create-mechanic/function-templates/d3-image/index.js
+++ b/packages/create-mechanic/function-templates/d3-image/index.js
@@ -7,29 +7,40 @@ export const handler = ({ inputs, mechanic }) => {
   const radius = ((height / 2) * radiusPercentage) / 100;
   const angle = Math.random() * 360;
 
-  let SVG = d3.create("svg")
-  .attr("width", width).attr("height", height)
+  const SVG = d3.create("svg")
+    .attr("width", width)
+    .attr("height", height)
+    .attr("viewBox", [0, 0, width, height])
+    .attr("style", "max-width: 100%; height: auto; height: intrinsic;");
 
-  let arc1 = d3.arc().innerRadius(0).outerRadius(radius)
-  .startAngle(0).endAngle(2*Math.PI/2);
+  const arc1 = d3.arc()
+    .innerRadius(0)
+    .outerRadius(radius)
+    .startAngle(0)
+    .endAngle(2*Math.PI/2);
 
-  let arc2 = d3.arc().innerRadius(0).outerRadius(radius)
-  .startAngle(2*Math.PI/2).endAngle(4*Math.PI/2);
+  const arc2 = d3.arc()
+    .innerRadius(0)
+    .outerRadius(radius)
+    .startAngle(2*Math.PI/2)
+    .endAngle(4*Math.PI/2);
 
   SVG.append("path")
-  .attr("transform", `translate(${center[0]},${center[1]}), rotate(${angle})`)
-  .attr("d", arc1).attr("fill", color1)
+    .attr("transform", `translate(${center[0]},${center[1]}), rotate(${angle})`)
+    .attr("d", arc1)
+    .attr("fill", color1)
 
   SVG.append("path")
-  .attr("transform", `translate(${center[0]},${center[1]}), rotate(${angle})`)
-  .attr("d", arc2).attr("fill", color2)
+    .attr("transform", `translate(${center[0]},${center[1]}), rotate(${angle})`)
+    .attr("d", arc2)
+    .attr("fill", color2)
 
   SVG.append("text")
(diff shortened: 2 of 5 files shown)


packages/create-mechanic/function-templates/svgjs-image/index.js
```diff
@@ -0,0 +1,74 @@
+import { SVG, extend as SVGextend, Element as SVGElement } from '@svgdotjs/svg.js'
```
fdoflorenzano, line 1: Is this the same dependency listed in the dependencies JSON? It looks like a different name. I tried running it and it can't find that module.

fdoflorenzano, line 1: Is this the same dependency listed in the dependencies JSON? It looks like a different name. I tried running it and it can't find that module.

packages/create-mechanic/function-templates/d3-image/dependencies.json
```diff
@@ -0,0 +1,6 @@
+{
+  "dependencies": {
+    "@mechanic-design/engine-svg": "^2.0.0-beta.7",
+    "d3": "^7.4.4",
```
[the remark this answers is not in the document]
fdoflorenzano: Looks fixed now!

packages/create-mechanic/function-templates/svgjs-image/dependencies.json
```diff
@@ -0,0 +1,6 @@
+{
+  "dependencies": {
+    "@mechanic-design/engine-svg": "^2.0.0-beta.7",
+    "svg.js": "^2.7.1",
```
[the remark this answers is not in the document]
fdoflorenzano: The trailing comma is fixed!

packages/create-mechanic/function-templates/svgjs-image/dependencies.json
```diff
@@ -0,0 +1,6 @@
+{
+  "dependencies": {
+    "@mechanic-design/engine-svg": "^2.0.0-beta.7",
+    "@svgdotjs/svg.js" : "^2.7.1"
```
fdoflorenzano, line 4: Tried installing it, but it threw error: `No matching version found for @svgdotjs/svg.js@^2.7.1`. I see you only changed the package name, but not the wanted version. Is it the same one?

fdoflorenzano (changes requested):
Found another little issue.

packages/create-mechanic/function-templates/svgjs-image/index.js
fdoflorenzano: It matches!

packages/create-mechanic/function-templates/svgjs-image/dependencies.json
fdoflorenzano: Now it works!

packages/create-mechanic/function-templates/d3-image/index.js
```diff
@@ -0,0 +1,82 @@
...
+  const angle = Math.random() * 360;
+
+  let SVG = d3.create("svg")
+  .attr("width", width).attr("height", height)
+
+  let arc1 = d3.arc().innerRadius(0).outerRadius(radius)
+  .startAngle(0).endAngle(2*Math.PI/2);
+
+  let arc2 = d3.arc().innerRadius(0).outerRadius(radius)
+  .startAngle(2*Math.PI/2).endAngle(4*Math.PI/2);
```
fdoflorenzano, line 17: Shouldn't these use `const` ? Is there something I'm missing where `let` is necessary?

packages/create-mechanic/function-templates/d3-image/index.js
```diff
@@ -0,0 +1,82 @@
...
+
+  SVG.append("path")
+  .attr("transform", `translate(${center[0]},${center[1]}), rotate(${angle})`)
+  .attr("d", arc1).attr("fill", color1)
+
+  SVG.append("path")
+  .attr("transform", `translate(${center[0]},${center[1]}), rotate(${angle})`)
+  .attr("d", arc2).attr("fill", color2)
```
fdoflorenzano, line 25: I do feel weird having D3 code not formatted in the "D3 style".

I mean like this:
```js
const svg = d3.create("svg")
    .attr("width", width)
    .attr("height", height)
    .attr("viewBox", [0, 0, width, height])
    .attr("style", "max-width: 100%; height: auto; height: intrinsic;");
```

It's not completely necessary really, this is more of an observation. What do you think?

fdoflorenzano (commented):
It works now! I do have some comments on the functions themselves. Let me know what you think!

packages/create-mechanic/function-templates/d3-image/index.js
fdoflorenzano: @munusshih looks like you changed this! it looks great!

packages/create-mechanic/function-templates/d3-image/index.js
fdoflorenzano: `const`s everywhere! nice!

packages/create-mechanic/function-templates/svgjs-image/index.js
```diff
@@ -0,0 +1,74 @@
...
+
+  const center = [width / 2, height / 2];
+  const radius = Math.max(radiusPercentage / 100, 0.1);
+  const angle = Math.random() * 360;
+
+  let draw = SVG().size(width, height)
+  , rawSVG = '<svg id="Layer_1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 90 90" ><defs></defs><path class="cls-1" d="M90,45A45,45,0,0,0,0,45Z"/></svg>'
+  , arcGroup = draw.group()
+  , arc1 = arcGroup.group().svg(rawSVG)
+  , arc2 = arcGroup.group().svg(rawSVG)
```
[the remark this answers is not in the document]
fdoflorenzano: @munusshih Looks like this hasn't been addressed. Do you have any thoughts about it?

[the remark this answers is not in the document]
fdoflorenzano: > @fdoflorenzano yes! Did you see my above comments?

I didn't 😂 .

I see now, it's an interesting use case. In that case, I suggest two things:
- Move the `rawSvg` var definition outside of the function, after the imports. Since it's a constant value, it's a bit of a mouthful of a sentence in the middle of the code. 
- Before the `rawSvg` definition, I would add a comment that explains it's usage. Maybe something like "this was exported from ..." "It can be replaced with another path export!" Or something like that. So that it's clearer on it's intention.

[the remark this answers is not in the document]
munusshih: > I didn't 😂 . 
That's fair cause it was pending. 😂 I forgot to submit them.
The suggestions sound great! I will implement them now.

fdoflorenzano (approved):
Amazing! Thank you so much for this! Do you want to do the honors and merge?

munusshih:
Sure!

PR #138, opened 2022-08-05
Fixes on fixes for v2.0.0-beta.9

fdoflorenzano:
### Added

- NPM package version shield on main README
- Docs version shield on main README
- PRs welcome shield on main README
- Added gif image header on main README

### Fixed

- Create CLI doesn't try on it's own to use yarn if possible, it asks if try to use npm or yarn
- Static folder gets uploaded again (I broke it in previous fix, ups)
- Corrects wrong name for "slider" property of number input in their definition
- Adds missing "multiple" property of image input in their definition
- Tweaks size of thumbnails in image input
- Resets cursor style on disabled image inputs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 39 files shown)


bravomartin (approved):
nice! 💐

PR #140, opened 2022-09-01
[139] Unique param storage key based on input configuration

### LINKED ISSUE #139: [v2.0.0-beta.9] Validation error when an input's type is changed and it conflicts with user's locally cached value
### Describe the bug
If an input's type is changed to a less permissive type (ex. color → text with specified options) the local value cached in the user's browser will throw an error due to failed validation.

### To Reproduce
1. Publish a function with a color input
2. Navigate to the published function and make a selection for that input
3. Change the color input type to "text" with several options
4. Publish a new version of the function to the same domain
5. Navigate to the same function in the same browser
6. You'll see something like "MechanicError: Error is thrown during input X value validation: Supplied Input value (#fafafa) is not available in options: Red, Yellow, Green"

### Expected behavior
I expect a function to ignore users' stale values for an input when the type is changed

### Screenshots
Screenshot of function throwing a MechanicError

### Desktop (please complete the following information):
 - OS: macOS Monteray
 - Browser: Chrome/Safari
 - Version: Latest

### Additional context
Changing the key value for the input avoids this scenario, but solutions like encoding the input type into the inputs name (themeText, themeColor) get cumbersome and are prone to mistakes.

Hashing individual input objects to create a unique ID (ala CSS modules) could be one way to ensure a stale value isn't used.

lnolte:
## Context
As described in #139 changes to the input's configuration of a function can cause bugs when the previously cached input values aren't compatible with the new input's configuration anymore.

## What this does
- adds a utility method that generates a hash from the string representation of an object
- uses this method to generate a hash based on a functions current input configuration
- adds that hash to that function's localStorage key, so it becomes `df_functionName_hash`
  - this makes sure that any changes to the input's config of a function will cause mechanic to create a fresh params cache

Closes #139

### DIFF
--- a/packages/core/app/components/utils/useValues.js
+++ b/packages/core/app/components/utils/useValues.js
@@ -143,6 +143,26 @@ const cleanValues = (object, reference) => {
   return object;
 };
 
+/**
+ * Turns the input object into a hash that can be used as
+ * a name for storing the latest values in localStorage.
+ *
+ * @param{object} inputs
+ * @returns{string}
+ */
+const generateHashFromInputsObject = inputs => {
+  const inputsAsString = stringify(inputs);
+
+  let hash = 0;
+  if (inputsAsString.length === 0) return hash;
+  for (let i = 0; i < inputsAsString.length; i++) {
+    const char = inputsAsString.charCodeAt(i);
+    hash = (hash << 5) - hash + char;
+    hash |= 0;
+  }
+  return hash.toString();
+};
+
 const useValues = (functionName, functionInputs, presets) => {
   const clean = useCallback(object => cleanValues(object, functionInputs), [functionInputs]);
   const initialValue = useMemo(() => {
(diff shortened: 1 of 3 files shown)


packages/core/app/components/utils/useValues.js
```diff
@@ -143,6 +143,26 @@ const cleanValues = (object, reference) => {
...
+  let hash = 0;
+  if (inputsAsString.length === 0) return hash;
+  for (let i = 0; i < inputsAsString.length; i++) {
+    const char = inputsAsString.charCodeAt(i);
+    hash = (hash << 5) - hash + char;
+    hash |= 0;
+  }
+  return hash.toString();
+};
+
```
runemadsen, line 165: I'm not smart enough to figure out if this can result in a super long has if the inputs are long? Should we use something that creates short hashes of e.g. 10-12 characters?

runemadsen (approved):
I left one question but everything works for me!

packages/core/app/components/utils/useValues.js
lnolte: The bitshift in line 160 should take care of the value never getting too long as any bits shifted of are discarded. But as this is also very old stuff from Uni for I added a substring to the end to be safe 😊

runemadsen:
Excellent! I think this is ready to go.

PR #145, opened 2022-09-02
opentype.js rendering for `dsi-logo-maker`

runemadsen:
This PR updates all design functions in `dsi-logo-maker` in the following way:

- They now use `opentype.js` for loading a `eot` version of F Grotesk and render text to `<path />` elements or canvas. This makes the output print-ready without the need for outlining fonts.
- They now all have string inputs with options for choosing the `thin` or `regular` version of F Grotesk


packages/dsi-logo-maker/functions/logoAnimatedSVG/index.js
```diff
@@ -7,7 +7,7 @@ import {
   getIndexModule,
 } from "../../utils/blocks";
 import { Unit } from "../../utils/blocks-components";
-import { useDrawLoop } from "../../utils/drawLoopHook";
+import { useDrawLoop } from "../../utils/hooks";
```
lnolte, line 10: I think the font loading is missing in this function :blush:

packages/dsi-logo-maker/functions/textAndImageCanvas/index.js
```diff
@@ -21,9 +23,10 @@ export const handler = ({ inputs, mechanic }) => {
   const words = text.split(" ").map((s) => s.toUpperCase());
   const colors = getColors("Custom Colors", null, colorsString.split(","));
   const height = Math.floor((width / ratio) * rows);
+  const font = await loadOpentypeFont(fontMode);
 
   const blockGeometry = computeBlockGeometry(width, height, rows, cols);
-  const baseBricks = computeBaseBricks(words, blockGeometry.fontSize);
+  const baseBricks = computeBaseBricks(words, blockGeometry.fontSize, font);
```
lnolte, line 29: This is probably unrelated to your changes, but the image input does not work for me on this one.

packages/dsi-logo-maker/functions/textAndImageSVG/index.js
```diff
@@ -25,39 +27,41 @@ export const handler = ({ inputs, mechanic }) => {
...
       reader.onload = function () {
         setHref(reader.result);
       };
-
       reader.onerror = function () {
         console.error(reader.error);
       };
     }
-  }, []);
+  }, [font, image]);
```
lnolte, line 43: Same here, clicking the image input doesn't open a file picker. As mentioned before, probably unrelated to your changes.

I'll look into what's causing this and open an issue for it :blush:

lnolte (changes requested):
This is looking amazing! I just found the font loading missing from one function 😊

packages/dsi-logo-maker/functions/textAndImageSVG/index.js
lnolte: Done #148

packages/dsi-logo-maker/functions/logoAnimatedSVG/index.js
runemadsen: Wooops!

packages/dsi-logo-maker/functions/textAndImageCanvas/index.js
runemadsen: Hmmm, weird. It was working for me when I tested it!

fdoflorenzano (approved):
Other than the missing updating for `logoAnimatedSVG/index.js` that Lucas pointed out and the font issue, it's looking good :)

runemadsen:
Thanks for the feedback @lnolte and @fdoflorenzano! I just pushed a fix so the `logoAnimatedSVG` function works with the font selector too. It's running a little choppy tho, and I can't really figure out why. I think we generally need to re-think the way we animate with React and provide some built-in hooks that people can use. I'll be happy to think about this for a bit, but for now I was just trying to fix the choppy logo and couldn't really understand how the `props` to the `Unit` component work together for the animation state. @fdoflorenzano do you mind taking a look? I'd also be happy to jump on a call :)

runemadsen:
Nice! I'm all for this!

runemadsen:
Another thing I was thinking in relation to `fps` was adding a small timeline scrubber for animations. It's a little hard since we don't know the duration of the animation, but perhaps it could be a small timeline with the ability to set the time you want to jump to. It would be a big help when working with animations.

runemadsen:
Let's solve the FPS problem when we merge the new `drawLoop` functionality. Getting this into `main` :)

PR #146, opened 2022-09-02
[WIP] Add export density to canvas renderer

lnolte:
- Makes canvas previews crisp on retina displays
- Allows users to choose the desired pixel density before exporting

### DIFF
--- a/packages/dsi-logo-maker/functions/textAnimatedCanvas/index.js
+++ b/packages/dsi-logo-maker/functions/textAnimatedCanvas/index.js
@@ -1,13 +1,13 @@
-import { getColors } from "../../utils/graphics";
+import { getColors } from '../../utils/graphics';
 import {
   computeBaseBricks,
   computeBlockGeometry,
   precomputeBlocks,
   getIndexModule,
-} from "../../utils/blocks";
-import { drawBlock } from "../../utils/blocks-canvas";
+} from '../../utils/blocks';
+import { drawBlock } from '../../utils/blocks-canvas';
 
-export const handler = ({ inputs, mechanic }) => {
+export const handler = ({ inputs, mechanic, getCanvas }) => {
   const {
     width,
     ratio,
@@ -20,20 +20,17 @@ export const handler = ({ inputs, mechanic }) => {
     loops,
   } = inputs;
 
-  const words = text.split(" ").map((s) => s.toUpperCase());
+  const words = text.split(' ').map((s) => s.toUpperCase());
   const height = Math.floor((width / ratio) * rows);
 
-  const colors = getColors("Custom Colors", null, colorsString.split(","));
+  const { canvas, ctx } = getCanvas({ height });
+
+  const colors = getColors('Custom Colors', null, colorsString.split(','));
   const blockGeometry = computeBlockGeometry(width, height, rows, cols);
   const baseBricks = computeBaseBricks(words, blockGeometry.fontSize);
 
   const blocksByIndex = precomputeBlocks(blockGeometry, baseBricks);
   const position = { x: 0, y: 0 };
-
-  const canvas = document.createElement("canvas");
-  canvas.width = width;
-  canvas.height = height;
-  const ctx = canvas.getContext("2d");
 
   const draw = () => {
     const brickOffset =
@@ -78,12 +75,12 @@ export const handler = ({ inputs, mechanic }) => {
 
 export const inputs = {
   width: {
-    type: "number",
+    type: 'number',
     default: 500,
     min: 100,
   },
   ratio: {
-    type: "number",
+    type: 'number',
     default: 9,
     max: 20,
     slider: true,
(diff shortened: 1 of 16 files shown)


runemadsen:
I was just thinking about this, and I think this is the right implementation. We should make sure that the `getCanvas` function throws an error if it doesn't have the needed `width` and `height`, whether that comes from inputs, the `getCanvas` arguments, or both. A nice error message with a call for the user to add the dimensions would be very helpful.

lnolte:
> I was just thinking about this, and I think this is the right implementation. We should make sure that the `getCanvas` function throws an error if it doesn't have the needed `width` and `height`, whether that comes from inputs, the `getCanvas` arguments, or both. A nice error message with a call for the user to add the dimensions would be very helpful.

I had the same notion and added some first drafts for error messages. https://github.com/designsystemsinternational/mechanic/pull/146/files#diff-cea56d01b8f19fd6436a19f2b4bbd9d2258c6edbeef3df917df2026fcf1ed520R82 :blush:

packages/engine-canvas/index.js
```diff
@@ -33,14 +50,79 @@ export const run = (functionName, func, values, config) => {
...
+    getCanvas: ({ width = null, height = null } = {}) => {
+      const dimensions = {
+        width: width || values.width || null,
+        height: height || values.height || null,
+      };
+
+      if (!dimensions.width && !dimensions.height) {
+        throw new Error(`No width and height values were provided to the canvas.
+
+You can set a widht and height values by passing them the getCanvas function:
```
fdoflorenzano, line 84:
```suggestion
You can set a width and height values by passing them the getCanvas function:
```

packages/engine-canvas/index.js
```diff
@@ -33,14 +50,79 @@ export const run = (functionName, func, values, config) => {
...
+
+        onDone(c);
+      },
       state: mechanic.functionState,
       setState: onSetState,
     },
+    getCanvas: ({ width = null, height = null } = {}) => {
+      const dimensions = {
+        width: width || values.width || null,
+        height: height || values.height || null,
```
fdoflorenzano, line 78:
```suggestion
        width: width ?? values.width ?? null,
        height: height ?? values.height ?? null,
```

packages/engine-canvas/index.js
```diff
@@ -1,14 +1,31 @@
...
+    : mechanic.exportDensity;
+
+  const preparedCanvas = document.createElement('canvas');
+  const preparedCanvasContext = preparedCanvas.getContext('2d');
+
+  if (values.width) preparedCanvas.width = values.width;
+  if (values.height) preparedCanvas.height = values.height;
+
+  preparedCanvas.style.transform = `scale${1 / canvasDensity}`;
+  preparedCanvasContext.scale(canvasDensity, canvasDensity);
```
fdoflorenzano, line 27: Is this canvas being used at all?

packages/dsi-logo-maker/functions/textAnimatedCanvas/index.js
```diff
@@ -1,13 +1,13 @@
-import { getColors } from "../../utils/graphics";
+import { getColors } from '../../utils/graphics';
```
fdoflorenzano, line 1: I think the only canvas design function missing is `textCanvas`

packages/engine-canvas/index.js
```diff
@@ -33,14 +50,79 @@ export const run = (functionName, func, values, config) => {
...
+  ...
+`);
+        }
+
+        onDone(c);
+      },
       state: mechanic.functionState,
       setState: onSetState,
     },
+    getCanvas: ({ width = null, height = null } = {}) => {
```
fdoflorenzano, line 75: I think it's a bit odd from the interface that we ask for the canvas and context from this function and then are supposed to give it back through `frame` and `done`. Would it make sense to keep a reference inside the engine and then use it when frame and done get called?

packages/engine-canvas/index.js
lnolte: Oh, good catch. That's leftover. Removed it

packages/engine-canvas/index.js
lnolte: Oh, I like it. I tried it and it's very nice. The only situation this is problematic is, when you also want to pass a `name` to `onDone`. So I changed `done` to accept an object as arguments. Let me know what you think.

packages/dsi-logo-maker/functions/textAnimatedCanvas/index.js
lnolte: :+1:

packages/engine-canvas/index.js
fdoflorenzano: I do like it! It does made me realize there's no mention in the documentation about the name argument for `done`. Maybe this is an opportunity to streamline arguments for the `frame` and `done` functions and leave all of them as single argument functions on all engines. Can be done in a separate PR tho.

runemadsen: I'm all in for this!

PR #154, opened 2022-09-20
fix: add property sorting for storage hashing

fdoflorenzano:
The tweak introduced here is adding key sorting to the key hash generation for input storage. Since the hash is computed from a string representation of a copy of the inputs export object, the order of the keys become important to generate the hash. This made the updated setup (introduced in https://github.com/designsystemsinternational/mechanic/pull/140) create a new storage even when changing the order of keys in the input, or even keys inside the values for inputs, which should (I think) preserve the storage for that definition.

### DIFF
--- a/packages/core/app/components/utils/useValues.js
+++ b/packages/core/app/components/utils/useValues.js
@@ -23,7 +23,9 @@ const copySerializable = obj => {
     }
   } else {
     copy = {};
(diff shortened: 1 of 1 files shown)


lnolte (approved):
💯

runemadsen (approved):
👍

PR #156, opened 2022-09-27
Adding decision document around new animation api

runemadsen:
You can read a nicer layout by [visiting the markdown file directly](https://github.com/designsystemsinternational/mechanic/blob/decision-new-animation-api/decisions/01-new-animation-api.md).

### DIFF
--- a/decisions/01-new-animation-api.md
+++ b/decisions/01-new-animation-api.md
@@ -1,0 +1,208 @@
+# Title
+
+## Status
+
+Proposed
+
+## Context
+
+Mechanic currently has support for animations. When setting `animation: true` in a design function, it receives the `frame()` and `done` callbacks that can be used to record individual frames into a movie file and finish the movie file all together. However, the framework doesn't come with any built-in tooling for running a draw loop inside your design function, so each template and example ships with its own helper code to facilitate this. For `engine-react`, this is a `useDrawLoop` hook that will re-render the React component 60 frames per second until the `done` callback is called. For the `engine-canvas`, we set up a `drawFrame` function that gets called with `requestAnimationFrame`.
+
+For the `2.0` release, we want to improve this setup to create a standardized animation API that cleans up the animated design functions by moving some logic into the Mechanic core, while also giving users the ability to write their own animation code if they choose so. Choosing a standardized animation API is a bit tricky with Mechanic, since it integrates with different kinds of frameworks and web technologies, and each of these frameworks have their own ways of doing things.
+
+- **p5.js** has its own animation api with the `setup` and `draw` functions. So `engine-p5` will most likely just want to use the `frame` and `done` callbacks as it currently does. Furthermore, this frame-based animation style encourages the use of global variables to hold the cumulated state of the animation.
+- **React** does not ship with any animation API, so users can either just re-render their component once per frame (which is a bit slow) or use something like `react-spring` to perform animations in a more event-driven fashion (which is faster because of the `react-spring` `animated` components that bypass the react rendering tree). Furthermore, React components are most of the time pure functions, which is a benefit for the timeline functionality mentioned below.
+- `engine-svg` and `engine-canvas` have no best practices since these are just slim wrappers around the HTML5 elements.
+
+One important aspect of this new animation API is that we want the ability to show a timeline scrubber in the Mechanic UI to give users ofan easy way to preview a specific frame of their design function output. This is much easier for frameworks that encourage functional design functions such as React and harder for frameworks such as p5.js. This timeline functionality is not described in this decision, but will build upon any decisions made in this proposal.
+
+## Decision
+
+We propose a new animation API where design functions default to a frame-based approach. If a design function is animated, it will automatically get called continously in a pace determined by the `frameRate` setting, until it calls the `done` function to stop the animation. This means that design functions are expected to be pure functions and not rely on global state to render the given frame. This in turn makes it possible for the Mechanic UI to provide a timeline that will update the design function with the given `frameCount` and result in a preview for that specific frame. Note that this new animation API does not impose any rules about the duration of the animation. This is always controlled by calling `done` at some point during the animation lifecycle.
+
+## Implementation Details
+
+The new animation API comes with a few changes to the settings:
+
+- The `animated` setting is renamed to `mode` (default: `static`) and the options are `static` (for static images), `animation` (for the new animation API) and `customAnimation` (for a user to implement their own animation code using the `frame` and `done` callbacks).
+- `frameRate` (default: `60`) is a number that can be used to change the number of frames per second that the design function is called in `animation` mode.
+
+Since the new animation API encourages pure functions, the engines now ship with a `memo` function that can be used to cache things across function calls. Engine-specific helpers such as a `useCanvas` function will also be rewritten to create the canvas element on the first call and return it on following calls.
+
+The argument syntax is also changing slightly by placing the `frame` and `done` functions as root properties of the design function argument object. The `frame` function will only get passed to the design function in `customAnimation` mode.
+
+Here's a look at what this new animation API will look like for each engine.
+
+### `engine-canvas`
+
+```js
+export const handler = async ({ inputs, frameCount, done, useCanvas }) => {
+  const canvas = useCanvas(inputs.width, inputs.height);
+  if (frameCount >= 100) {
+    done(canvas);
+  }
+  // drawing code
+  return canvas;
+};
+
+export const settings = {
+  engine: require('@mechanic-design/engine-canvas'),
+  mode: 'animation'
+};
+```
+
+```js
+export const handler = async ({ inputs, frame, done, useCanvas }) => {
+  const canvas = useCanvas(inputs.width, inputs.height);
+  let frameCount = 0;
+  const myDrawLoop = () => {
+    frameCount++;
+    // drawing code
+    if (frameCount >= 100) {
+      done(canvas);
+    } else {
+      frame(canvas);
+      window.requestAnimationFrame(myDrawLoop);
+    }
+  };
+  myDrawLoop();
+};
+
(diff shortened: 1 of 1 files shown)


decisions/01-new-animation-api.md
```diff
@@ -0,0 +1,208 @@
...
+- **React** does not ship with any animation API, so users can either just re-render their component once per frame (which is a bit slow) or use something like `react-spring` to perform animations in a more event-driven fashion (which is faster because of the `react-spring` `animated` components that bypass the react rendering tree). Furthermore, React components are most of the time pure functions, which is a benefit for the timeline functionality mentioned below.
+- `engine-svg` and `engine-canvas` have no best practices since these are just slim wrappers around the HTML5 elements.
+
+One important aspect of this new animation API is that we want the ability to show a timeline scrubber in the Mechanic UI to give users ofan easy way to preview a specific frame of their design function output. This is much easier for frameworks that encourage functional design functions such as React and harder for frameworks such as p5.js. This timeline functionality is not described in this decision, but will build upon any decisions made in this proposal.
```
lnolte, line 17:
```suggestion
One important aspect of this new animation API is that we want the ability to show a timeline scrubber in the Mechanic UI to give users an easy way to preview a specific frame of their design function output. This is much easier for frameworks that encourage functional design functions such as React and harder for frameworks such as p5.js. This timeline functionality is not described in this decision, but will build upon any decisions made in this proposal.
```

decisions/01-new-animation-api.md
```diff
@@ -0,0 +1,208 @@
...
+  engine: require('@mechanic-design/engine-canvas'),
+  mode: 'animation'
+};
+```
+
+```js
+export const handler = async ({ inputs, frame, done, useCanvas }) => {
+  const canvas = useCanvas(inputs.width, inputs.height);
+  let frameCount = 0;
+  const myDrawLoop = () => {
```
lnolte, line 58: I like this. Maybe the drawloop could still just come as a helper from the core.

decisions/01-new-animation-api.md
```diff
@@ -0,0 +1,208 @@
...
+};
+```
+
+### `engine-svg`
+
+```js
+export const handler = async ({ inputs, frameCount, done }) => {
+  // drawing code
+  if (frameCount >= 100) {
+    done(svgString);
```
lnolte, line 83: Just thinking about this. This forces users to keep `svgString` in a variable so it can either be returned or be send to `done`. While this isn't necessarily a problem as you have to the same using `frame` and `done` currently I think it might feel weird as `return` encourages to just build the return value there (at least to me).

So maybe the API could be simplified, that in `animation` mode `done` just flips a boolean in the engine that tells the engine to stop rendering after the next return value is received.

decisions/01-new-animation-api.md
```diff
@@ -0,0 +1,208 @@
...
+  return <div></div>;
+};
+
+export const settings = {
+  engine: require('@mechanic-design/engine-react'),
+  mode: 'animation'
+};
+```
+
+The example above re-renders the component for every frame. If a user wants to use something like `react-spring` to speed up animation while keeping the frame-based approach, it could be done like this:
```
lnolte, line 132: Just leaving this here to consider when implementing this:

I think the way the react engine currently works this would cause a re-render of everything every time, because the engine would call React DOM's `render` over and over again. When implementing this using the the new animation approach the engine should probably be re-build to render a wrapping component only once and then use a drawLoop within that parent component to re-render to function over and over again.

decisions/01-new-animation-api.md
```diff
@@ -0,0 +1,208 @@
...
+
+export const settings = {
+  engine: require('@mechanic-design/engine-react'),
+  mode: 'customAnimation'
+};
+```
+
+### `engine-p5`
+
+For `engine-p5`, `mode` can only be set to `static` or `customAnimation`. Or perhaps we default both `animation` and `customAnimation` to the code below?
```
lnolte, line 175: I like this. Maybe a descriptive info could be printed on the console, to educate developers about why the new animation isn't possible using p5

decisions/01-new-animation-api.md
```diff
@@ -0,0 +1,208 @@
...
+
+### Positive Consequences
+
+The benefits for most users is that they will get an animation API that gets out of their way. The templates and examples for animated design functions will be a lot slimmer, and it will provide us with an API that we can build nice things on top of.
+
+### Negative Consequences
+
+The main disadvantage I can see is:
+
+- Users who are not comfortable writing pure functions are now defaulted into this, and if they opt out, they need to write their own animation code.
```
lnolte, line 208: I think this disadvantage can be mitigated by shipping animation loop utils for each engine that a user can just import into their function.

lnolte (commented):
I think this is a very nice proposal already. My only real remark would be that we should ship drawLoop utils a user can import when they are using the new `customAnimation` behavior. So they only need to write their own drawLoops if they really want to.

decisions/01-new-animation-api.md
runemadsen: I was actually just thinking about this. I think there's some weirdness around whether the user needs to pass something to the callbacks or not. So I like the idea of `animation` removing the need to do that and `customAnimation` requiring you to do that. We just need to keep in mind the difference that in `animation` mode, the `done` function will be called before the return, thus the last frame will not be recorded as a part of the animation, whereas it will be when you pass the string to the `done` function in `customAnimation`.

decisions/01-new-animation-api.md
runemadsen: Yes, I agree. We need to take advantage of the React render loop, so the component should not render from scratch using `render` every time. good point!

decisions/01-new-animation-api.md
runemadsen: Yes!

decisions/01-new-animation-api.md
runemadsen: Agreed!

decisions/01-new-animation-api.md
runemadsen: I agree. That would be very nice!

runemadsen:
Great! After thinking about it, I'm not a big fan of the `customAnimation` mode name. I'm wondering if we can rename it to something that makes more sense, but I don't really have a good idea yet. `static` and `animation` still seem good to me.

decisions/01-new-animation-api.md
runemadsen: That, or we change it so the `done` function never receives anything and always ends the animation immediately. That way, the signature for the `frame` and the `done` function will stay consistent, and the different modes are simply about whether you use the `frame` function or not. I think I like that slightly better.

lnolte: Agreed

decisions/01-new-animation-api.md
```diff
@@ -0,0 +1,208 @@
...
+
+The argument syntax is also changing slightly by placing the `frame` and `done` functions as root properties of the design function argument object. The `frame` function will only get passed to the design function in `customAnimation` mode.
+
+Here's a look at what this new animation API will look like for each engine.
+
+### `engine-canvas`
+
+```js
+export const handler = async ({ inputs, frameCount, done, useCanvas }) => {
+  const canvas = useCanvas(inputs.width, inputs.height);
```
fdoflorenzano, line 40: I'm not sure I love the "use" prefix here. To me that's a React specific convention, and this specific example isn't React. I would try to expose arguments and functions that live in the same universe at the current engine.

decisions/01-new-animation-api.md
```diff
@@ -0,0 +1,208 @@
...
+  if (frameCount >= 100) {
+    done(canvas);
```
fdoflorenzano, line 42: Maybe this is very engine specific, but does done have to receive the canvas? what happens with the one returned right after? I guess that's something that sticks out to me in this new way of writing design functions.

EDIT: now I see you commented on this too in another section. I agree with something like that should also be applied here.

fdoflorenzano:
> Great! After thinking about it, I'm not a big fan of the `customAnimation` mode name. I'm wondering if we can rename it to something that makes more sense, but I don't really have a good idea yet. `static` and `animation` still seem good to me.

Since they are strings anyways, I was thinking on going with `"custom-animation"` really, but yeah maybe it's not great. I'm honestly fine with it. Maybe a separate setting only applicable for `"animated"`, that ahs values:`"frame-based"` and `"event-based"`, so we start this animation basis language in the code too.

runemadsen:
Thanks @fdoflorenzano! I think both of those could work. If we keep the single `mode` setting, then perhaps `static`, `animation` and `animation-custom`?

lnolte:
I updated #152 with a draft implementation showing the new engine behavior for the two new animation modes.

A few observations:
- Rewriting the frame and done methods as discussed above removes the need for treating them mutually exclusive in an if-else-block.
  - you only need to wrap done in an if clause, but can call frame however you want, as done will make sure to exit the function after the next call of frame
  - this logic is needed to correctly capture the last frame
- Refactoring settings from `animated?: booelan` to `mode?: 'static' | 'animation' | 'animation-custom'` is a breaking change to existing design functions
  - how do we want to communicate this? Is it docs only? Or should we add a `checkDeprecations` method to the mechanic core, that can run on initialization and check for any deprecated settings being present?

runemadsen:
@fdoflorenzano I have edited the doc and I think it's ready for you to take a look :)

fdoflorenzano:
@runemadsen I like it! A couple of thoughts came to mind:

1. Is there an animated setting or similar in the settings of a design function? I didn't see any in the examples. Maybe having an example for the static case will help this feel complete.
2. For an animation DF, if I understood correctly, the api let's the user not use the `drawLoop`helper, right? Maybe adding an example of that too would help exemplify the flexibility aspect of the setup.

PR #158, opened 2022-10-26
Adding Decision Document around Timeline

lnolte:
I gave it a first pass to write down the results of our timeline brainstorm from last week.

### DIFF
--- a/decisions/02-timeline-functionality.md
+++ b/decisions/02-timeline-functionality.md
@@ -1,0 +1,69 @@
+# Timeline control for animated design functions
+
+## Status
+
+Proposed
+
+## Context
+
+When fine tweaking values for an animated design function developers might find
+themselves in a situation where they just a need a single frame of the animation
+for reference. They want to be able to update an input value or a constant in
+code and judge the visual effect of this change based on a single frame.
+
+We had this experience when using Mechanic to develop the title and credit
+sequences for a movie.
+
+Using the current setup there are ways of manually adding this behavior, but
+they require developers to work with their own drawloops, add the needed input
+and take care of the conditional rendering (animate vs. still frame) themselves.
+
+This document proposes a timeline input as part of mechanic core. It is intended
+as an opt-in developer feature that's most useful in purely frame based
+animations. It makes heavy use of the [proposed](https://github.com/designsystemsinternational/mechanic/pull/156) new [Animation API](https://github.com/designsystemsinternational/mechanic/pull/157).
+
+## Decision
+
+We propose a timeline input that developers can opt into for animated design
+functions. To do so they'd have to set the `showTimeline` property of their
+function's settings to `true`.
+
+By default the timeline renders with two additional number inputs for the min
+frame number and the max frame number. The timeline scrubber interpolates
+between these two values. This makes sure a developer can zoom in and zoom out
+of the animation area they need.
+
+Alternatively developers can hardcode the maximum number of frames for the
+timeline by passing the `timelineMaxFrames` setting. If this is set, the timline
+input will just show the scrubber and interpolate between `0 ` and the value of
+`timelineMaxFrames`.
+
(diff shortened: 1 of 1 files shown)


decisions/02-timeline-functionality.md
```diff
@@ -0,0 +1,69 @@
...
+between these two values. This makes sure a developer can zoom in and zoom out
+of the animation area they need.
+
+Alternatively developers can hardcode the maximum number of frames for the
+timeline by passing the `timelineMaxFrames` setting. If this is set, the timline
+input will just show the scrubber and interpolate between `0 ` and the value of
+`timelineMaxFrames`.
+
+The timeline is a development feature only. When building your design tool for
+deployment it is not exported.
```
runemadsen, line 42: I'm not entirely sure about this. I can easily see users wanting to preview certain frames of their exports based on their input settings.

runemadsen (changes requested):
I like it! I think it's simple and nice. One thing I think we need to consider is play/stop functionality and how this interacts with the timeline. Right now the animation just runs from the beginning whenever you update an input value. But if the timeline is set, I guess it needs to just show that frame? And what decides whether it's playing or not? I think the timeline functionality should solve this too, perhaps with some sort of autoplay feature or actual controls in the timeline.

decisions/02-timeline-functionality.md
```diff
@@ -0,0 +1,69 @@
...
+## Decision
+
+We propose a timeline input that developers can opt into for animated design
+functions. To do so they'd have to set the `showTimeline` property of their
+function's settings to `true`.
+
+By default the timeline renders with two additional number inputs for the min
+frame number and the max frame number. The timeline scrubber interpolates
+between these two values. This makes sure a developer can zoom in and zoom out
+of the animation area they need.
```
fdoflorenzano, line 34: Why should there be a min input and no setting to hardcode it? Is there a situation really where we need a min frame? isn't it always 0? Removing it also would help with the confusion of range export.

decisions/02-timeline-functionality.md
fdoflorenzano: Agreed!

fdoflorenzano (commented):
Looking good! I felt a comment regarding the min input, and I agree with @runemadsen about the pause and play functionality. I do think it's weird to have inconsistent autoplay depending on the timeline.

I'm thinking: timeline has the play/pause and allows it to play from any starting point thanks to the range, and no timeline just always auto-plays.

decisions/02-timeline-functionality.md
lnolte: Using `min` and `max` inputs allows a user to zoom to specifc ranges of frames if they want to.

WIth the hardcoded value it is assumed a user will always want the entire animation as the range of the timeline (thus `0` is implied as min value in that case).

decisions/02-timeline-functionality.md
lnolte: Ok, what would that mean for the export then? Would you be able to export only the range of the timeline then? Does setting the timeline to one frame allow you to export a single frame then?

runemadsen: I would argue that the setting of the timeline has no affect on the export. The export exports the whole thing no matter the setting on the timeline. This is also how it works in After Effects. You can zoom, but an export is the whole thing.

fdoflorenzano: Same here! I understood Rune's initial comment about the deployment of the tool. In a dev or prod scenario the timeline could be useful.

decisions/02-timeline-functionality.md
fdoflorenzano: Interesting. I didn't get that implication from the setting of the `timelineMaxFrames` setting. Maybe it's a naming thing.

Issue #159, opened 2022-11-03
Enable increasing value by 10x the step when pressing shift  + arrow


bravomartin:
**Is your feature request related to something something in Mechanic? Please describe.**
When changing values in the numeric inputs, it would be nice to enable faster editing the values.

**Is your feature request related to a problem? Please describe.**

It becomes hard to change the values quickly with the keyboard and it doesn't align with the behavior of other design tools.
I think this will also make the need for rewriting values altogether (with the know issues it has) less frequent 

**Describe the solution you'd like**

Make it so that when the user is pressing the `shift` key, the value changes 10× the specified step of the input.

**Describe alternatives you've considered**
 none

PR #160, opened 2022-11-08
159 number input improvements

lnolte:
## What this does

- This addresses #159, holding shift will cause the number input to increment by step * 10 using the arrow keys
- This also addresses the issue observed by Rune that is related to the value being clamped inside the number component
  - you're now allowed to type a value that is out of min/max range
  - the number input component just won't emit this value, so you don't get a re-render
  - the number input component will display an error message

## Pending changes

- I think the min/max validation shouldn't happen inside the number component but rather at the parent level
- However, I believe this is a bigger refactor, as we'd need to refactor the parent, so it stops event emission if there is a validation error somewhere along its children
- The current behavior (with the custom validators a user can provide) is that the function will re-render, even if some input field has a validation error

## Next steps

Before tackling the potential refactor, I'd like @fdoflorenzano take on the potential refactor.

I'd also like @runemadsen to check, if this new number input behavior (allowing out-of-range values but erroring) fixes the issue he mentioned.

Also @bravomartin should check if the 10x step is what he had in mind.

### DIFF
--- a/packages/ui-components/src/input/NumberInput.js
+++ b/packages/ui-components/src/input/NumberInput.js
@@ -1,8 +1,10 @@
-import React, { useRef, useState, useMemo } from "react";
+import React, { useRef, useState, useMemo, useEffect } from "react";
 import PropTypes from "prop-types";
 import classnames from "classnames";
 import { uid } from "../uid.js";
 import { Invalid } from "../icons/index.js";
+
+import { useShiftKey } from "../util/useShiftKey.js";
 
 import * as commonCss from "../common.module.css";
 import * as css from "./NumberInput.module.css";
@@ -32,8 +34,12 @@ export const NumberInput = props => {
   } = props;
   const [internalValue, setInternalValue] = useState(value);
   const rangeError = useRef(null);
+  const isShiftPressed = useShiftKey();
   const [focus, setFocus] = useState(false);
 
+  // TODO: This validation needs to be moved outside the component up to the
+  // parent. We just need to make sure to stop event propagation on error at the
+  // parent level, so we do not cause a Mechanic Error to be thrown.
   const isInRange = v => {
     if (min !== undefined && v < min) {
       rangeError.current = `Must be at least ${min}`;
@@ -47,6 +53,7 @@ export const NumberInput = props => {
     return true;
   };
 
+  // TODO: Refactor, so this lives inside the parent
   const computedInvalid = useMemo(() => {
     return invalid || rangeError.current !== null;
   }, [invalid, rangeError.current]);
@@ -54,6 +61,14 @@ export const NumberInput = props => {
   const computedError = useMemo(() => {
     return error || rangeError.current;
   }, [error, rangeError.current]);
+
+  // When shift is pressed, we want to increment/decrement by 10x the step
+  const computedStep = useMemo(() => {
+    if (step !== undefined) {
+      return step * (isShiftPressed ? 10 : 1);
+    }
+    return isShiftPressed ? 10 : 1;
+  }, [step, isShiftPressed]);
 
   const handleOnChange = event => {
     const { name, value } = event.target;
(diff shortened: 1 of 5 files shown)


runemadsen:
Yezzz, this solves my problem!

runemadsen (approved):

packages/ui-components/src/input/NumberInput.js
```diff
@@ -30,13 +32,52 @@ export const NumberInput = props => {
...
+    return error || rangeError.current;
+  }, [error, rangeError.current]);
+
+  // When shift is pressed, we want to increment/decrement by 10x the step
+  const computedStep = useMemo(() => {
+    if (step !== undefined) {
+      return step * (isShiftPressed ? 10 : 1);
+    }
+    return isShiftPressed ? 10 : 1;
+  }, [step, isShiftPressed]);
```
fdoflorenzano, line 71: Is useMemo worth it for any of these values?

packages/ui-components/src/input/NumberInput.js
```diff
@@ -116,22 +157,24 @@ export const NumberInput = props => {
             aria-describedby={`error-${id}`}
             aria-invalid={invalid}
```
fdoflorenzano, line 158: Should this also use `computedInvalid`?

fdoflorenzano (commented):
Thanks for this Lucas! The step part works great!

I left some comments regarding the implementation of the other part of the PR, the min/max thing, but I found a bigger problem than those specific ones.

This change is a bit tricky. This implementation uses internal state for the input to try to tackle it, which transforms this component to suddenly not being a controlled component. Initially the components value will match the one it gets from the app, but as soon as it rendered, they are separate states. Even though for valid values, the value will get propagated back to the app, the state managed by the input is independent. And as every uncontrolled component, synchronization is tough. In the preview you can experience a un-sync if you toggle a number component, and then setting a preset. The app will update the DF with the new value, but the input value won't change. 

So it's weird, cause you need an internal value that gets updated when the given value also updates. But at the same time, allowing the internal value to update itself without updating the outside one. 

I guess a useEffect call could be done where it updates the internal value when the prop value changes?? but I'm not sure of the whole implications of that.

runemadsen:
This is a good catch @fdoflorenzano and I haven't thought about that. I've managed these situations before, and it always gets super messy. Let me think about this and let's talk about it Monday.

lnolte:
> I guess a useEffect call could be done where it updates the internal value when the prop value changes?? but I'm not sure of the whole implications of that.

I think I'd prefer keeping the Input components purely controlled and move the validation to the parent layer, where it could get merged with any user-provided custom validation. That would probably mean moving the validation logic to either the Input.js wrapper or the Sidebar in core, do you agree or do you think there's a better place for it?

I also noticed, that if a user provided validation fails this doesn't currently stop the function from re-rendering. If moving all validation to the parent I'd like to change that behavior, so that any failing validation (no matter if built-in or user-provided) stops the function from re-rendering.

If this proposal sounds good to you, I'm more than happy to give the needed refactor a shot on this PR :blush:

runemadsen:
This is sorta what I was thinking too. I'm still not sure how it would solve our number input problem, but if @fdoflorenzano agrees, I think it would be great to give it a try and see if it can be solved this way.

fdoflorenzano:
> I think I'd prefer keeping the Input components purely controlled and move the validation to the parent layer, where it could get merged with any user-provided custom validation.

Yeah, that also sounds like a good road to go in to me.



> If moving all validation to the parent I'd like to change that behavior, so that any failing validation (no matter if built-in or user-provided) stops the function from re-rendering.

This is where I wonder if we'll get to the same problem we are having now. I guess it makes sense for the validation to work this way, but I personally never thought about it that way. 

Fine by me if you take a stab at this! If it gets tough tho, I would recommend separating that task into it's own PR so we can get the shit scaling in.

lnolte:
I had another look at this, I'll do the refactor to the validation logic on another PR, as it feels like a bigger refactor. I'll close this for now.

I've cherry-picked the shift-step onto #167 for isolated review, rework and potential merge :blush:

PR #161, opened 2022-11-11
All pretty pretty!

fdoflorenzano:
Did a little choring formatting all files with the recent addition of a prettier config. Other recent PRs had a lot of formatting changes that made it a bit difficult to follow which were format changes and which were for the PR. Let's get this over with!

This PR:
- Adds prettier commands at the root of the monorepo
- Runs prettier formatting across the repo
- Replaces setting `jsxBracketSameLine` (cause it's [deprecated](https://prettier.io/docs/en/options.html#deprecated-jsx-brackets)) with `bracketSameLine`. Because of the deprecation, I had a weird situation where the prettier command didn't leave the bracket in the same line, but the prettier from VSCode did! I changed it to `false`cause I noticed in Lucas' PRs that the bracket was left in another line, and I like it better I think :) .

### DIFF
--- a/.github/ISSUE_TEMPLATE/---bug-report.md
+++ b/.github/ISSUE_TEMPLATE/---bug-report.md
@@ -1,17 +1,17 @@
 ---
 name: "\U0001F41B Bug report"
 about: Report an issue with Mechanic.
-title: ''
-labels: 'bug:bug:'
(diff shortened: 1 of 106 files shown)


runemadsen (approved):
Excellent! Let's get this in :)

PR #162, opened 2022-11-11
fix: dsi-logo-maker/textAndImageSVG

fdoflorenzano:
While checking #157, I noticed the textAndImageSVG function in dsi-logo-maker didn't export. This was because of a separate issue, where a done call would be done when the function renders null. This fixes that little issue.


runemadsen (approved):
Good catch! Let's meeeerge

PR #167, opened 2022-11-25
Allow 10x steps by holding shift on number input

### LINKED ISSUE #159: Enable increasing value by 10x the step when pressing shift  + arrow

lnolte:
## Context
Bringing this change over from #160, so it can be reviewed and potentially merged without being blocked by the bigger refactor to the validation flow proposed in #160.

## What this does
- This addresses #159, holding shift will cause the number input to increment by step * 10 using the arrow keys

Closes #159

### DIFF
--- a/packages/ui-components/CHANGELOG.md
+++ b/packages/ui-components/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+- Holding down shift allows increases the step of the numeric input by 10
+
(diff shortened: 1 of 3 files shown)


fdoflorenzano (approved):
👏🏽

PR #170, opened 2023-02-03
fix: port option in serve command didn't override config

### LINKED ISSUE #166: mechanic serve --port does not override mechanic.config.js
**Describe the bug**
The `mechanic serve` command has an option name `--port` that states in the help:
```
      --port        Custom port to serve app                     [default: 3000]
```
so by running `mechanic serve --port 3030` I would expect to be served over the 3030 port but it's override by the value that is defined in `mechanic.config.js`

**To Reproduce**

1. create a new mechanic project
2. run `mechanic serve --port 3030`

**Expected behavior**

I do expect that the `--port` overrides the .config file and or remove the option entirely to not have confusion on duplicate entry point for the same option. Even if the very preferable way would be to be able to define it via ENV variables, since in different environments the port could change, and should not leverage on a source tracked file.

**Additional context**
I'm super sorry reporting many things, just put them here because I belive that the project has a huge potential, and a reminder to myself, for future contributions (:

fdoflorenzano:
Fixes #166


runemadsen (approved):
Looks good to me 👍

packages/core/src/commands/serve.cjs
```diff
@@ -38,7 +38,7 @@ const command = async argv => {
 
   spinner.start("Starting off server...");
   // Set port and express server
-  portfinder.basePort = config.port != null ? config.port : argv.port;
+  portfinder.basePort = argv.port !== null ? argv.port : config.port;
```
lucasdinonolte, line 41: I noticed that for both the serve and the dev command `3000` is passed in as the default in `yargs`. 

I think this would mean that this ternary now has the opposite effect: It will always serve on port 3000 (or any other port you send into the CLI) and always ignore the port specified in your config.

lucasdinonolte (commented):
I left a question regarding the default port always set in `yargs`.

packages/core/src/commands/serve.cjs
fdoflorenzano: Good catch. Tho it's not opposite to what this PR wants, that is to make this behavior consistent between the two commands and take the argv port before config. But yes, given the way it's setup now, the config gets completely ignored (at least in this section).

The config and it's usage isn't really that well realized or though through. It's an idea it made sense at the beginning of developing Mechanic but haven't come back to it or giving it much protagonism (it's not even documented I believe). I propose separating the config issue from this and discuss how to use it and it can be handle in relation to commands.

lucasdinonolte: Ok got it! Makes sense, I'm all for merging then and (maybe) addressing the "issue" when we give the role of the config more thought.

PR #171, opened 2023-02-03
fix: when using templates or examples take name arg

### LINKED ISSUE #165: create command with --template does not allow function names
**Describe the bug**
When run `yarn new --template react-image my_design_function` the latest positional argument is not considered.
This creates always a function in `functions/react-image`

**To Reproduce**
1. create mechanic project
2. run `yarn new --template react-image my_design_function`

**Expected behavior**
I do expect to be able to pass a custom function design name ;p

fdoflorenzano:
Fixed #165


fdoflorenzano:
@runemadsen I realized I missed something in the new project section to make it consistent with this fix, and also used a local definition of nullish coalescing as the rest of these scripts.

runemadsen (approved):

PR #173, opened 2023-05-03
Little create-mechanic fix

fdoflorenzano:
There was a little line break and `>` character missing in the create-mechanic flow when a user installs after creating the design function. This adds them!


runemadsen (approved):
Lovely!

PR #174, opened 2023-05-03
feat(core): adds hideFunction setting

fdoflorenzano:
This was something I've encountered a couple times myself while doing prototypes in mechanic. I create a bunch of design functions, trying different stuff, until I get to a final definite design function. To hide all other early explorations, I have to manually move everything out of the `functions/` folder. But adding this setting, let's us just hide with a setting quickly, and keep track of them even in HMR mode.


runemadsen (approved):
This will be super useful. Thanks for the PR! Approved!

PR #175, opened 2023-05-05
create-mechanic improvements

fdoflorenzano:
A bunch of pending improvements for create-mechanic! (Update: also contains a fix for dev and build commands)

- `dev` and `build` prompts have the functions and inputs folders paths as options to specify when running, but it was set up to have default values which would fallback to those values at the prompt level. That meant that the script didn't actually get whenever the command didn't get the option set up at all. Because of this, the script wouldn't take the value specified in the project's config at all. I removed this default fallback at the prompt level, so the config is taken into account. The eventual fallback still exists if the config doesn't specify folder paths.
- The most important that has been pending, this tweaks how design function templates and examples are managed, so that it also supports copying functions with custom inputs! The directory structure was changed to make this easier, so you'll see a lot of moved files. Each design function base has a directory with a `function/` folder (design function source code), a `dependencies.json` file (with a list of dependencies for the DF), and may have an `inputs/` folder with any custom inputs the DF uses.
- The previous change allows us to finally add the Adaptive Grid example @munusshih made a while ago! It used a custom input, and it will soon be able to replicate it for mechanic users across! Thanks again for this @munusshih! The other examples you made will be added as soon as possible too.
- Now create will also add an .gitignore file to a new project. As well as try to initialize a new git repository. This is a little improvemene that will make creating DFs and posting to GitHub more easy and quick.
- Added a little detail for the installing section. The cli asks users if they want to install dependencies, and now it will also list which dependencies are being installed.
- Added some line breaks across different sections to make the CLI output more readable.

Screenshot 2023-05-05 at 13 15 00
Screenshot 2023-05-05 at 13 15 18

### DIFF
--- a/packages/create-mechanic/function-examples/adaptive-grid/dependencies.json
+++ b/packages/create-mechanic/function-examples/adaptive-grid/dependencies.json
@@ -1,5 +1,6 @@
 {
   "dependencies": {
-    "@mechanic-design/engine-react": "^2.0.0-beta.10"
+    "@mechanic-design/engine-react": "^2.0.0-beta.11",
+    "@mechanic-design/ui-components": "^2.0.0-beta.11"
   }
 }
--- a/packages/create-mechanic/function-examples/adaptive-grid/inputs/group/style.module.css
+++ b/packages/create-mechanic/function-examples/adaptive-grid/inputs/group/style.module.css
@@ -1,11 +1,10 @@
 .root {
   margin: 1em 0;
-}
-
-.toggle {
-  margin: 1em 0;
+  padding: 1em;
+  border: 1px solid currentColor;
+  border-radius: var(--mechanic-input-height-half);
 }
 
 .inputs {
-  padding-left: 1.5em;
+  margin-top: 1em;
 }
--- a/packages/create-mechanic/function-examples/adaptive-grid/inputs/group/index.js
+++ b/packages/create-mechanic/function-examples/adaptive-grid/inputs/group/index.js
@@ -34,7 +34,6 @@ export const Input = props => {
   return (
     <div className={css.root}>
       <BooleanInput
-        className={css.toggle}
         label={inputDef.label}
         name="show"
         value={show}


packages/create-mechanic/function-examples/adaptive-grid/function/assets/FuturaStd-Bold.otf
lucasdinonolte, line 1: This should probably replaced by an open source font.

Hate to recommend poppins, but I think we can get away visually with replacing it with Poppins :blush:

Other alternatives could be

- https://fonts.google.com/specimen/Jost
- https://fonts.google.com/specimen/League+Spartan

munusshih: Oh yah tots agree. I remember I was only using Futura, because that was what was used on the original logo. But we should definitely replace it by an open source font.

Screenshot 2023-05-10 at 2 48 09 PM

fdoflorenzano: Sorry that this took so long for me to get back to! But I finally did! 

Great catch @lucasdinonolte! Funnily enough, the one example selected to add here didn't actually use Futura. So for now it's enough to just remove the file and unused declaration.

Tho in the near future where we do add the other examples, we would need to replace Futura. @munusshih Do you have any font replacement preference for it? Lucas linked to some nice alternatives.

fdoflorenzano:
I also went ahead and cleaned up some other unused CSS declarations, and formatted some files that weren't before.

packages/create-mechanic/function-examples/adaptive-grid/inputs/group/style.module.css
```diff
@@ -0,0 +1,7 @@
+.root {
+  margin: 1em 0;
+}
+
+.toggle {
+  margin: 1em 0;
+}
```
lucasdinonolte, line 7: Should we box the group's content off a bit more visually? I had a hard time understanding what the custom component does initially.

packages/create-mechanic/function-examples/adaptive-grid/function/index.js
```diff
@@ -0,0 +1,411 @@
...
+  },
+  Ticket: {
+    width: 394,
+    height: 1126
+  }
+};
+
+export const settings = {
+  engine: require("@mechanic-design/engine-react"),
+  showMultipleExports: true
```
lucasdinonolte, line 410: When exporting this as a PNG the image that's rendered in the preview by default doesn't show up. When I upload a custom image it works.
grid-2025-05-06-21-09 (1)
grid-2025-05-06-21-09

packages/create-mechanic/function-examples/adaptive-grid/function/index.js
```diff
@@ -0,0 +1,411 @@
...
+          <p>{textThree}</p>
+        </div>
+      </Column>
+
+      <Column
+        width={threeWidth}
+        height={fullHeight}
+        x={chooseX[2] + borderRatio * 0.5}
+        y={borderRatio}
+        showGrid={grid}
```
lucasdinonolte, line 223: I think this could be a cool use case for the `isPreview` argument that's passed to a design function now, as — if I understand correctly — the grid should probably not be included in the export.

lucasdinonolte (commented):
Very nice, left some comments. I'm not sure how much the added example function is meant to be reviewed, so feel free to ignore the comments made on the example function.

The only thing I really found is that I think this is missing to pull the inputs path from the mechanic config, so potentially copying any custom inputs to an unexpected folder.

packages/create-mechanic/new-function.js
```diff
@@ -156,16 +178,39 @@ const generateFunctionTemplate = async (
...
         JSON.stringify(packageObj, null, 2)
       );
     })()
   ]);
+
+  // Add any custom inputs
+  const templateHasInputs = await fs.pathExists(inputsSrcDic);
+  const generatedCustomInputs = { tried: false, success: false };
+  if (templateHasInputs) {
+    const projectInputDir = path.join(projectDir, "inputs");
```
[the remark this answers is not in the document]
fdoflorenzano: Great catch! Fixed this and did some tweaks on the script to reflect the awareness of the folders. 

This made me realized too that dev and build didn't take the config paths into account because the values always defaulted in the prompt level, so the fallback to config never made it there. I added a fix for this.

packages/create-mechanic/function-examples/adaptive-grid/function/index.js
fdoflorenzano: Good idea! Added :)

packages/create-mechanic/function-examples/adaptive-grid/function/index.js
fdoflorenzano: Thanks URL loading. Switched it to load the image directly.

packages/create-mechanic/function-examples/adaptive-grid/inputs/group/style.module.css
fdoflorenzano: Any suggestions? I added a bit of padding so the nested relationship is more clear.

munusshih:
So happy to see that you guys are still working on this :)! Let me know if there's anything I can do to help.

fdoflorenzano:
> So happy to see that you guys are still working on this :)! Let me know if there's anything I can do to help.

omg thanks @munusshih! Sorry it took us this long to continue working on this. We've been busy, but now we're pushing to get it out :) I think we've got it covered, but we'll let you know if anything comes up.

packages/create-mechanic/function-examples/adaptive-grid/dependencies.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "@mechanic-design/engine-react": "^2.0.0-beta.10"
```
lucasdinonolte, line 3: Do we need to add `@mechanic-design/ui-components` here as well?

I tested this locally by adding this function to the DSI Logo maker and had to install `@mechanic-design/ui-components` manually, as the app would crash otherwise.

lucasdinonolte (changes requested):
Shaping up nicely! Just found one small thing around a missing dependency of the new function-example.

packages/create-mechanic/function-examples/adaptive-grid/inputs/group/style.module.css
fdoflorenzano: Thanks for this! I did a tweak to be closer to this.

packages/create-mechanic/function-examples/adaptive-grid/dependencies.json
fdoflorenzano: Great catch! I replicated it and go to the error too. Should be fixed now.

lucasdinonolte (approved):
Amaaaaaaazing

Issue #129, opened 2022-05-04
Custom input CSS infiltrates into SVG outputs


fdoflorenzano:
**Describe the bug**
When custom inputs are used in mechanic project, their CSS also infiltrates into a design function's iframe because of dependencies, and therefore SVG outputs also get injected their CSS. In some cases, that CSS isn't automatically removed. I specifically had this with the `ui-components` CSS definitions, and had SVG outputs that triple in weight because of these.

A specific setting is introduced (`ignoreStyles`) in #128 to patch this, but that also ignores the design function's CSS. So currently having only the design function's CSS but not any from custom inputs is a bit tricky.

**Expected behavior**
SVG outputs not to contain custom input's CSS automatically.

**Desktop (please complete the following information):**
 - OS: macOS Monterrey 12.3.1
 - Browser: Firefox 99.0.1

PR #176, opened 2023-05-11
Removing unsued input styles in function exports

fdoflorenzano:
This is a first direction of fixing an old bug introduced with custom inputs.

This was first noted when working with custom inputs that used some of Mechanic's ui-components. Some SVG based design functions included in the project that included the custom inputs, would fail at export (either PNG or SVG) with the following message:

`Uncaught (in promise) Error: Pseudo-elements are not supported by css-select. `

Through some digging, it was found that this is thrown by SVGO (that uses css-select) when trying to compute optimizations over the resulting SVG. It threw that because it turns out CSS meant for input elements (like text field, number field, etc...) got bundled into the SVG being exported, and those included pseudo-elements.

This happens because a section was added when adding CSS support for SVG, that clones all style in the design functions iframe into the SVG, so that exported SVG actually included styles defined in the design function. But, it wasn't considered at the time that this would carry over any other styles that may not even be for the design function, like the ones from custom inputs. Custom inputs style is there in the first place because Mechanic adds dependencies to the script run in the iframe that imports the inputs code, so it can run validations and other value things, but therefore also adds any style associated with inputs.

This problem was first posted as #129. I go back and forth on whether if it's an actual problem that the input's CSS gets to the iframe. What is a problem for sure is that it makes it to the export. 

This draft includes a potential fix focusing on the export step. Search through style rules that are included in the iframe, and skipping any that has rules that don't match to anything in the export or raise error when searching. Those error might be because of badly written CSS or unsupported selector (like pseudo-elements).

Adds two dependencies that are already included by SVGO: css-select (to query rule selectors over the SVG) and css-tree (to parse and write CSS strings).

Let me know what you think! I will open a new fresh PR with cleaner commits if we decide to take it in.


fdoflorenzano:
> That being said, I would prefer that we prevent the CSS from being in iframe in the first place, since it doesn't really seem like it's needed there.

Yeah, that's the part where I kept going back on forth on. Cause in a way what's behind that is just webpack adds it following the dependency chain. If the custom input imports whatever CSS for no reason related to the input validations, that would also be added, and it would make sense (in the sense that, it's a dependency, so it should be there).



> Can you point me towards the exact place where the CSS is being injected?

Sure, it's a bit of a maze at this point. The code that gets run directly in the iframe is this [section](https://github.com/designsystemsinternational/mechanic/blob/51e1c72df72dcc92e56ef598364ad3dcb0c3ec80/packages/core/src/commands/utils.cjs#L67-L75).
  - Which imports in the first line the results generated [here](https://github.com/designsystemsinternational/mechanic/blob/main/packages/core/src/commands/utils.cjs#L106-L108). That goes through the input definitions that checks that stuff makes sense (just in the input defs). That script is [here](https://github.com/designsystemsinternational/mechanic/blob/main/packages/core/src/input-set-up.js).
  - Then, the iframe script, runs the setUp function defined [here](https://github.com/designsystemsinternational/mechanic/blob/main/packages/core/src/function-set-up.js). Which in some other stuff, would run the input validation functions for the values passed from the UI. 
  
 We could try to figure out how to refactor this so that the iframe is able to call the validations for the passed values, _without_  directly importing that codebase.

fdoflorenzano:
@runemadsen I tested this again too. Even though the base error is a bit hard to recreate (I could do it directly with a simple custom input), the base functionality that removes unnecessary CSS from the inputs does work.

runemadsen (approved):
Nice, seems to work as described! I'm okay getting this in and then having us rethink the whole iframe setup for future releases. If we do rewrite mechanic as a bunch of util functions, I'm not convinced that we need the iframe at all. But that's for another time :)

fdoflorenzano:
> Nice, seems to work as described! I'm okay getting this in and then having us rethink the whole iframe setup for future releases. If we do rewrite mechanic as a bunch of util functions, I'm not convinced that we need the iframe at all. But that's for another time :)

That's fair. I also feel it could generally re-thought. I'll add the changelogs for this and merge.

PR #179, opened 2023-06-05
Disable export button in UI while exporting

lucasdinonolte:
A tiny quality of life change to the UI.

- The `Export` and `Generate` buttons are disabled while an export is running to more clearly communicate to the user that something is happening
- This is done by allowing the UI to subscribe to events from the mechanic runtime
- This could thus easily be extended to emit events on every frame to present the user with a growing number or progress bar while exporting


fdoflorenzano (commented):
I love the idea, and the events approach is interesting to me. Some other features could be approached from this point.

Tho, I think it's not working, specifically the try to catch the startDownload moment. I ran it locally with a fresh new project, I export once and the button then remains disabled.

The approach here tries subscribing on the lastRun object, which already ran the engine, and therefore the download part. 

I feel like there's space here then to re-factor how the mechanic object run cycle behaves. Cause right now, it's the UI, saying to the iframe to run the engine, it's the engine who creates the mechanic object instance with all the values and other things, it validates things but also goes ahead and runs the DF. Then after that it returns to the UI the result of the run. I think there's space to reconsider how that life-cycle works, maybe the mechanic instance is begun in the UI with the information it has at the moment, there's time to subscribe to events before actually running the DF, and then telling the iframe to run. Also with the events idea, I feel like there's room to also map out the lifecycle as an internal state based lifecycle, which the UI responds to based on subscriptions.

Tho this might be a bit much before the 2.0.0 update. So maybe let's reconsider if any other "simpler" options come to mind.

lucasdinonolte:
@fdoflorenzano I gave this some more thought and I think I overall like the event emitter model, as it would theoretically also allow any external part to subscribe to mechanic events (if we ever want that). So I left the `on`, `off`, and `emit` methods in place and just added a way to pass a set of initial event listeners to the constructor via the `config`. This way I can make sure these listeners get always added to the runtime before any rendering happens. I'm still 100% with you on the idea of sketching out the full mechanic lifecycle and potentially refactor the responsibility of who's creating the mechanic instance in the future. So I tried to keep my changes a non-invasive as possible, to not block that road. Let me know what you think.

fdoflorenzano (approved):
I tried it again now and it worked great in image and video functions! Thanks for this :) 

I like the UX enhancement it brings, I could also see doing something similar for preview/generate button, and adding maybe spinners or what not for different loading states. This could be done separately though and have it in mind for a lifecycle review.

lucasdinonolte:
@runemadsen wanna have a look at this too before we merge it?

runemadsen (approved):
Love it!

PR #190, opened 2025-04-25
Pass isPreview boolean prop to design function

### LINKED ISSUE #189: Pass preview/export information to design function
For a small icon builder I was building with mechanic, I was rendering a grid of SVG rectangles as visual guides for the user when drawing in their icon. I only want those grid lines in preview, but not when exporting the icon to SVG.

I think it could be great if the `mechanic` object that's being passed to a design function could have a `isPreview` or `isExport` boolean property.

**Describe alternatives you've considered**
I've worked my way around using a boolean input, but that puts manual effort on the user, needing to make sure to hide the grid before exporting.

lucasdinonolte:
This PR changes all existing mechanic engines to pass the `isPreview` to the design function. It's added as a new prop and can be destructured in the design function’s handler.

This is useful if for example you want to render guidelines in preview, but don’t want to include them in the final export.

Closes #189

### DIFF
--- a/packages/engine-canvas/CHANGELOG.md
+++ b/packages/engine-canvas/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Expose `isPreview` to design function.
+
 ## 2.0.0-beta.0 - 2022-03-16
 
 ### Added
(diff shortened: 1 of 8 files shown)


lucasdinonolte:
Draft PR to reflect this change in the docs here: https://github.com/designsystemsinternational/mechanic.design/pull/38

runemadsen (approved):

fdoflorenzano (approved):
Tested with all engines and works great! Thanks for this :)

PR #194, opened 2025-07-08
fix: ui-components latest version in dependencies

fdoflorenzano:
I realized while testing `create-mechanic` locally that the version I added for `ui-components` in the last comment in #175 doesn't exists. This fixes it. It also adds some logs on the `updateVersions.js` script, that I used to fix this, just to make sure everything was working.


lucasdinonolte (approved):
LGTM

PR #197, opened 2026-01-27
Upgrade webpack and related dependencies

lucasdinonolte:
Another round of dependency updates, this time updating all dependencies from the broader "webpack group" (so babel, webpack itself, webpack plugins, etc.)

This is built on top of the react upgrade, to keep the diff small, but also to incrementally upgrade and make sure everything is compatible.

***

Package | Update Type
-- | --
@babel/cli | Minor
@babel/core | Minor
@babel/plugin-transform-runtime | Minor
@babel/preset-env | Minor
@babel/preset-react | Minor
@babel/runtime | Minor
babel-loader | 2 Major Versions
clean-webpack-plugin | 1 Major Version
copy-webpack-plugin | 3 Major Versions
css-loader | 1 Major Version
css-minimizer-webpack-plugin | 4 Major Versions
html-loader | 3 Major Versions
html-webpack-plugin | Minor
mini-css-extract-plugin | Minor
node-polyfill-webpack-plugin | 3 Major Versions
postcss-loader | 2 Major Versions
postcss-preset-env | 3 Major Versions (tried 4 major versions, but that dropped CJS support)
style-loader | 1 Major Version
terser-webpack-plugin | Minor
webpack | Minor
webpack-dev-middleware | 2 Major Versions
webpack-bundle-analyzer | 1 Major Version
webpack-dev-server | Minor
webpack-hot-middleware | Minor
loader-utils | Removed, as webpack 5+ loaders can use `this.getOptions()` instead
regenerator-runtime | Removed, as browser support for `async/await` and `yield` is good enough by now
*** 

So far it's looking good, but I'll need to test it more rigorously before putting it up for review.

Once this is done we should discuss how to best tackle the remaining dependencies, that don't really fit the "React" or "Webpack" groups. Those would be things like `svgo` in core, or `yargs` on cli. I think it'd probably make sense to go package by package once React and Webpack are out of the way.

***

Sources:
https://caniuse.com/es6-generators
https://caniuse.com/async-functions
https://github.com/webpack/loader-utils (repo is archived now, because this is native to webpack 5)
https://webpack.js.org/api/loaders/#thisgetoptionsschema


lucasdinonolte:
When initially testing this in a local mechanic project (using the `yalc` to link the packages) I was sometimes observing that some design functions would load themselves into the `iframe` including their sidebar, so you’d have mechanic inside of mechanic. I couldn’t really pin down the specific constellation that made this happen, and re-checking it just now also couldn’t reproduce it, so it might just have been something related to `yalc` as I was switching branches and versions of local mechanic quite frequently when testing initially.

@fdoflorenzano I think it’s still worthwhile to put extra attention to testing everything with a local mechanic project that’s not the DSI logo maker, as bumping webpack and friends really feels like a scary change.

fdoflorenzano (changes requested):
It's working great on my end! The only update I think is now needed is updating the "engines" in the corresponding packages to cover from `20.19.0`. By change I locally had `20.15.1`, got the warnings when installing, but then running dev threw errors. Updating to `20.19.0` fixed it. It comes from some of postcss dependency updated I believe.

lucasdinonolte:
> It's working great on my end! The only update I think is now needed is updating the "engines" in the corresponding packages to cover from `20.19.0`. By change I locally had `20.15.1`, got the warnings when installing, but then running dev threw errors. Updating to `20.19.0` fixed it. It comes from some of postcss dependency updated I believe.

This is a good find, newer node version just silently will allow to call `require` on an ESM file, while older versions of node do not allow it. `postcss-preset-env` dropped all CJS support moving to version 11 (which was released January of this year). The latest version of `postcss-preset-env@10` was also released this January and still has CJS support, given that we're not updating any of the ui component styles using new CSS features that would need transpilation I feel like going down to version 10 feels a little less like a sledge hammer than the node.js version bump.

I fully agree though, that on the next dependency upgrade PR we should look into which node.js versions we want to support. I'm happy to then upgrade postcss-preset-env to 11 again, if we decide to go with a version that will silently allow `require('/path.mjs')`.

As you'll see on the next PR this whole CJS vs ESM shift that happened since the creation of mechanic is a bigger issue holding us back when it comes to updating CLI dependencies, so keeping postcss-preset-env on a CJS compatible version feels in line with the other dependency upgrades to me. Happy to chat.

fdoflorenzano (approved):
Amazing! I think that's a good call, thanks for proposing this alternative. Tested with new projects and works nicely!

PR #199, opened 2026-02-24
Dependency updates (others)

lucasdinonolte:
After focussing on React and Webpack dependency upgrades in the previous PRs, this PR upgrades a bunch of the other dependencies, while also not updating all of them, see below for explanation:

## Updates done
Package | Update Type | Notes
-- | -- | --
express | Major | Simple usage; no breaking changes affecting us.
fs-extra | Major | API unchanged; drops support for Node 12 & 14.
semver | Minor | –
classnames | Minor | –
proptypes | Minor | –
portfinder | Minor | –
yargs | Minor | –
inquirer | Minor | –
chalk | Minor | –

## CLI Dependencies

A note on the CLI dependencies, a lot of them are way behind their latest version, but they all made the switch to ESM, while the mechanic CLI is still CJS. So without moving everything to ESM we're stuck with the version updates I've done, which I honestly don't know if it's worth it, considering we're already planning for Mechanic 3

Package | ESM Only From | Latest Version
-- | -- | --
yargs | v18 | 18
inquirer | v9 | 12
chalk | v5 | 5
ora | v6 | 9
execa | v6 | 9
get-port | v6 | 7

## Node.js Support

This PR also updates the `engine` field on all packages that will get directly installed into a user's projects to ensure they are running node.js version 20 or above. 20 is the oldest LTS version at the time of authoring this, with its EOL being scheduled for April 2026.

See: https://nodejs.org/en/about/previous-releases

It's worth noting, that depending on how the user has set up their npm config, an older version will not stop them from installing mechanic, but they will get a bunch of warnings when installing.

Screenshot 2026-03-12 at 9 56 38 AM

## Not Updated

- `svgo` comes with changes to how the plugin system works, so there's a high likelihood this will break the config we allow users to set in `optimize`. From a quick glance, there also seem to be no security issues on the version of `svgo` we're on, so leaving it as is should be fine
- `html-to-image` it's a visual package that would need heavy testing + going from HTML to PNG isn't a mechanic feature we really advertise in the docs, so I left this as is
- `lerna` see below

## A note on `lerna`

Lerna dropped support for the `bootstrap` command in more recent versions, so updating lerna would mean changing the monorepo setup quite a bit. Since we're already sketching out a version 3 of mechanic, where we'll probably move the monorepo to something more "native" like `npm workspaces` I didn't bother looking into the lerna upgrade path too much.

> We no longer include the bootstrap, add, and link commands by default. We strongly recommend using your package manager (npm, yarn, pnpm) for package management related concerns such as installing and linking dependencies.

## Maybe on Follow Up

- `p5` version used by `engine-p5`, though I would want to discuss internally if we stay on the `1.x` branch of p5 or go to `2.x`

## Other Observations

- I found that `react-color` seems to be no longer maintained (last commit was six years ago), while the color picker works just fine, I just want to raise this, that whenever we wanna tweak something around the color input would probably be a good time to search for a replacement

Depends on #197 being merged first


fdoflorenzano (approved):
Amazing Lucas! It all worked well on my end, and all the decisions described make sense to me. The only one I don't have an opinion a way or the other, is with P5. But it can be handled in another PR or here if discussed with Rune and you want to do it. 

Thanks again for all this work, truly.

PR #200, opened 2026-03-27
Upgrade p5 version in engine-p5

lucasdinonolte:
This PR updates the p5.js version used from 1 to 2.
- Updates the dependency in `engine-p5`
- Updates poster example, that uses `engine-p5`, to new version


runemadsen (approved):
Nice!

